PR #152, opened 2014-10-03
Use strongloop test MySQL server for testing

raymondfeng:
/to @bajtos or @ritch 

The PR allows we to use https://github.com/strongloop/strongloop-config for testing mysql.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "main": "app.js",
   "scripts": {
     "pretest": "test -z \"$JENKINS_HOME\" || (echo | node bin/setup-mysql.js)",
-    "test": "mocha"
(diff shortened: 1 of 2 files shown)


superkhau:
Closing this as rejected. @raymondfeng Are you keen on getting this in still? its been 2 years. I'm proposing we default to not using mysql in your unit tests (since they are really integration tests since they rely on third party such as MySQL). We should move those out into loopback-example-workspace which gets triggered downstream by cis-jenkins now and we can also use that example project to double up as acceptance/system tests @deepakrkris is already working on this with the workspace.next refac.

If you still feel differently, feel free to reopen.

Issue #170, opened 2014-12-02
Support async boot scripts OOTB


bajtos:
We need to rework `server/server.js` template to support async boot scripts.

```
boot(app, __dirname, function(err) {
  if (err) throw err;
  // start the server if `$ node server.js`
  if (require.main === module) {
    app.start();
  });
```

In order to support the scenario when the app is loaded via `require` but not started, e.g. in loopback-sdk-angular-cli, we need to modify loopback-boot to emit an event when the boot process is finished, e.g. `app.emit('ready')`.

**Tasks**

 - [x] https://github.com/strongloop/loopback-boot/pull/77 modify loopback-boot to emit `ready` event (or similar)
 - [x] https://github.com/strongloop/loopback-workspace/pull/172 rework `server/server.js` template to defer `app.start` until `boot(app)` is finished
 - [x] https://github.com/strongloop/generator-loopback/pull/74 fix `slc loopback:example` to use async boot script for generating models, see [example/templates/server/boot/create-sample-data.js](https://github.com/strongloop/generator-loopback/blob/master/example/templates/server/boot/create-sample-data.js)

chandadharap:
@bajtos The three tasks look check-marked as done?

bajtos:
> @chandadharap The three tasks look check-marked as done?

@superkhau ticked them off before his pull requests were landed. I removed the check-marks and added links to the relevant pull requests.

superkhau:
Close.

superkhau:
Done.

PR #172, opened 2014-12-02
Boot the app asynchronously

superkhau:
Connected to #170

### DIFF
--- a/templates/components/api-server/template/server/server.js
+++ b/templates/components/api-server/template/server/server.js
@@ -3,10 +3,6 @@ var boot = require('loopback-boot');
 
 var app = module.exports = loopback();
 
-// Bootstrap the application, configure models, datasources and middleware.
-// Sub-apps like REST API are mounted via boot scripts.
-boot(app, __dirname);
-
 app.start = function() {
   // start the web server
   return app.listen(function() {
@@ -15,7 +11,12 @@ app.start = function() {
   });
 };
 
-// start the server if `$ node server.js`
-if (require.main === module) {
-  app.start();
-}
+// Bootstrap the application, configure models, datasources and middleware.
+// Sub-apps like REST API are mounted via boot scripts.
+boot(app, __dirname, function(err) {
+  if (err) throw err;
+
+  // start the server if `$ node server.js`
+  if (require.main === module)
+    app.start();
+});
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -140,6 +140,19 @@ describe('end-to-end', function() {
         done(err);
       });
     });
+
+    it('emits the `booted` event when booting is complete', function(done) {
+      var src = FIXTURES + '/async.js';
+      var dest = SANDBOX + '/server/boot/async.js';
+      fs.copySync(src, dest);
+      delete require.cache[require.resolve(SANDBOX)];
+      var app = require(SANDBOX);
+      app.on('booted', function() {
+        expect(app.asyncBoot, 'app.asyncBoot').to.be.true();
+        done();
+      });
+      // the test will time out if `booted` is not emitted
+    });
   });
 
   describe('autoupdate', function() {
--- a/test/fixtures/async.js
+++ b/test/fixtures/async.js
@@ -1,0 +1,4 @@
+module.exports = function(app, cb) {
+  app.asyncBoot = typeof cb === 'function';
+  process.nextTick(cb);
+};
--- a/templates/components/api-server/component.js
+++ b/templates/components/api-server/component.js
@@ -14,7 +14,7 @@ template.package = {
     "compression": "^1.0.3",
     "errorhandler": "^1.1.1",
     "loopback": "^2.8.0",
-    "loopback-boot": "^2.4.0",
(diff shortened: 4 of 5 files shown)


superkhau:
@bajtos I notice `npm test` has failures (most likely due to https://github.com/strongloop/loopback-boot/issues/94 as the fail log messages look familiar)

>scaffold a new app

Should I copy the code over from https://github.com/strongloop/loopback-workspace/blob/master/test/support.js#L79-L85? I don't see any code in boot that scaffolds an app.

superkhau:
@bajtos Review please.

templates/components/api-server/template/server/server.js
```diff
@@ -3,19 +3,21 @@ var boot = require('loopback-boot');
...
 var app = module.exports = loopback();
 
-// Bootstrap the application, configure models, datasources and middleware.
-// Sub-apps like REST API are mounted via boot scripts.
-boot(app, __dirname);
-
 app.start = function() {
   // start the web server
   return app.listen(function() {
-    app.emit('started');
```
bajtos: We must preserve the event `started`, it's used by [server/boot/explorer.js](https://github.com/strongloop/loopback-workspace/blob/f031f73f00b903204a674fcfdb6570fa0bad88cc/templates/components/api-server/template/server/boot/explorer.js#L20-L26) to print a message with Explorer URL, but only after the app was started.

templates/components/api-server/template/server/server.js
```diff
@@ -3,19 +3,21 @@ var boot = require('loopback-boot');
...
-// start the server if `$ node server.js`
-if (require.main === module) {
-  app.start();
-}
+// Bootstrap the application, configure models, datasources and middleware.
+// Sub-apps like REST API are mounted via boot scripts.
+boot(app, __dirname, function(err) {
+  if (err) throw err;
+
+  app.emit('ready');
```
bajtos: An earlier comment says "This should be implemented by loopback-boot, please remove". Boot already emits `booted` event, this is what consumer should use.

test/end-to-end.js
```diff
@@ -140,6 +140,24 @@ describe('end-to-end', function() {
...
     });
+
+    it('emits the `ready` event when booting is complete', function() {
+      var src = FIXTURES + '/emit.js';
+      var dest = SANDBOX + '/server/boot/emit.js'
+      fs.copySync(src, dest);
+      delete require.cache[require.resolve(SANDBOX)];
+      var app = require(SANDBOX + '/server/server');
+      expect(app.readyEventEmitted).to.be.ok();
+    });
```
bajtos: This test is not necessary, the event `booted` is triggered by `loopback-boot`. 

However, if you feel it's valuable to write an integration test capturing the contract that a server scaffolded by the template will fire `booted` event, then I am ok with that too. Just rename `ready` to `booted` everywhere.

test/fixtures/async.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function(app, cb) {
+  // ensure boot script executes asynchronously
+  app.asyncBoot = typeof cb === 'function';
```
bajtos: This does not really verify that the script are executed asynchronously. You should check that the boot process waits until the async scripts finished.

I think you should rework the two new tests and merge them into one.

```js
// async.js
module.exports = function(app, cb) {
  app.asyncBoot = typeof cb === 'function';
  process.nextTick(cb);
}

// in the test
var app = require(...);
expect(app.booting, 'app.booting').to.be.true();
app.on('booted', function() {
  expect(app.asyncBoot, 'app.asyncBoot').to.be.true();
  done(); 
});
// the test will time out if `booted` is not emitted
```

Thoughts?

templates/components/api-server/template/server/server.js
superkhau: I see, its different from `booted`. Adding back. ;)

templates/components/api-server/template/server/server.js
superkhau: I forgot about the `booted` event, removing.

test/end-to-end.js
superkhau: Rebanubg `ready` to `booted`.

test/fixtures/async.js
superkhau: I agree. Merging into one simple script instead.

superkhau:
@slnode test please

superkhau:
@bajtos Review please.

bajtos:
One more thing to fix before you can merge this patch: The scaffolded version of loopback-boot (`2.4.0` ATM - [source](https://github.com/strongloop/loopback-workspace/blob/f031f73f00b903204a674fcfdb6570fa0bad88cc/templates/components/api-server/component.js#L17)) does not support async boot scripts. Please bump up the version to `^2.6.5`.

Note that the version of loopback-boot in the toplevel `package.json` must remain at 1.x, that's a completely different thing.

Other than that, the change LGTM. You can land it without further review after you upgrade loopback-boot.

bajtos:
Also please make sure the test is passing on Jenkins before merging.

PR #182, opened 2015-01-12
Fix status middleware

bajtos:
Before this change, the scaffolded application had status middleware
configured to handle all requests. As a result, it was handling unknown
URLs too.

This patch partially reverts 3071a26 and brings back the original server/boot/root.js script.

**Current version**

```
//server/middleware.json
{
  "routes": {
    "loopback#status": {
      "paths": "/"
    }
  }
}
```

**Upcoming version**
```
//server/boot/root.js
module.exports = function(server) {
  // Install a `/` route that returns server status
  var router = server.loopback.Router();
  router.get('/', server.loopback.status());
  server.use(router);
};
```

---

A part of strongloop/generator-loopback#80

/to @ritch or @raymondfeng please review

### DIFF
--- a/templates/components/api-server/template/server/boot/root.js
+++ b/templates/components/api-server/template/server/boot/root.js
@@ -1,0 +1,6 @@
+module.exports = function(server) {
+  // Install a `/` route that returns server status
+  var router = server.loopback.Router();
+  router.get('/', server.loopback.status());
+  server.use(router);
+};
--- a/templates/components/api-server/template/server/middleware.json
+++ b/templates/components/api-server/template/server/middleware.json
@@ -12,9 +12,6 @@
   "parse": {
   },
   "routes": {
-    "loopback#status": {
-      "paths": "/"
-    }
   },
   "files": {
   },
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -61,6 +61,13 @@ describe('end-to-end', function() {
         });
     });
 
+    it('provides status on the root url only', function(done) {
+      // See https://github.com/strongloop/generator-loopback/issues/80
+      request(app)
+        .get('/does-not-exist')
+        .expect(404, done);
+    });
+
     it('has authentication enabled', function(done) {
       request(app)
         .get('/api/users')
@@ -210,7 +217,7 @@ describe('end-to-end', function() {
 
   describe('discovery', function() {
     this.timeout(10000);
-    
+
     var connection;
     before(function(done) {
       connection = setupConnection(done);


raymondfeng:
The `scopeToGetRoot` is not very intuitive. Maybe a better way to fix this issue is to extend `paths` in middleware.json to allow methods too.

bajtos:
> The `scopeToGetRoot` is not very intuitive.

I agree.

> Maybe a better way to fix this issue is to extend `paths` in middleware.json to allow methods too.

That's IMO very un-express-like. It also does not solve this particular problem, which lies in fact that `loopback.status` was originally meant to be used as a route via `app.get`, not as a middleware mounted via `app.use`.

The difference is that middleware handles all sub-paths at a given endpoint, while middleware handles only the root path at the given endpoint. Example:

```js
app.use('/status', status()); // handles /status/foobar too
app.get('/status', status()); // does not handle /status/foobar
```

I don't think `middleware.json` should be used to build (compose) express routes.

If we cannot come up with a better name for the config option, then I can revert the change in loopback-workspace that moved registration `loopback.status` from `boot/root.js` to `middleware.json`.

Thoughts?

ritch:
Well... IMO `status()` is not a middleware at all. It is actually a route handler. Ideally a middleware does not have any internal routing and calls `next()`. Obviously there are exceptions to this (`static()` for example).

I'd rather see this as `app.get('/', ...)`. Perhaps like this...

```js
// boot/status.js
if(app.get('root status')) {
  app.get('/', function() {
    // ...
  });
}
```

Then we can enable and disable this via `server/config.json` and by deleting or adding the boot script. The downside is obviously ordering... Boot scripts do not guarantee order so that may be an issue if you want to use it alongside another handler.

bajtos:
@raymondfeng thoughts? @ritch is ok with `scopeToGetRoot` per his comment in strongloop/loopback#996

bajtos:
Ignore my comment above, I am going to delete it.

bajtos:
> Then we can enable and disable this via server/config.json

That's IMO overcomplication. I'll revert back to the solution we have before moving `status` to middleware.json:

https://github.com/strongloop/loopback-workspace/blob/290cfba4b40188036fc4a6af833be0abfe7400f1/templates/components/api-server/template/server/boot/root.js

```js
module.exports = function(server) {
  // Install a `/` route that returns server status
  var router = server.loopback.Router();
  router.get('/', server.loopback.status());
  server.use(router);
};
```

bajtos:
I have reworked the patch to partially revert 3071a26 and bring back the original `server/boot/root.js` script.

@raymondfeng @ritch PTAL

ritch:
:+1:

bajtos:
Before we can land and release this patch, `yo loopback:example` must be modified to support the old-now-new flavour again: https://github.com/strongloop/generator-loopback/pull/82

@chandadharap @dashby3000 When this patch is landed and released, the project structure scaffolded by `slc loopback` will becom slightly different. Let me know if you have any training that could be affected by this change, I can delay the release in such case.

**Current version**

```
//server/middleware.json
{
  "routes": {
    "loopback#status": {
      "paths": "/"
    }
  }
}
```

**Upcoming version**
```
//server/boot/root.js
module.exports = function(server) {
  // Install a `/` route that returns server status
  var router = server.loopback.Router();
  router.get('/', server.loopback.status());
  server.use(router);
};
```

bajtos:
Released as `v3.7.0`

Issue #181, opened 2015-01-08
nesting package.json files in LB project causes corruption when saving datasource


seanbrookes:
not sure if this is specific to datasource save flow in Arc but the use case is:

- I have a LB/Arc project.
- it has a 'sub project' for a node app running on a tessel device.
- I was trying to keep the tessel app code in a folder within my LB project so the code was all in one place 
eg:
root/
../client/
../server/
../tessel/
  - package.json
...
package.json

- when I try to save a data source definition I get an error message in the UI indicating the package.json is corrupted.  

It looks like LB is doing some kind of recursion through the project files and picking up both package.json files.

bajtos:
The fix is available in `loopback-workspace@3.6.5`.

@seanbrookes please verify. Don't forget to update the workspace module in your Arc installation first.

cgole:
I verified by adding a sub directory and an invalid package.json file. Worked as expected. loopback ignored the invalid package.json

cgole:
@sean is that  test case good enough to validate the fix?

seanbrookes:
you should have 2 valid packge.json files

one in the root of your project and another in a 'child' folder (think nested node.js apps)

PR #183, opened 2015-01-14
Fix package.json handling

bajtos:
- Remove PackageDefinition-related code from Facet, since PackageDefinition is always top-level, it is never associated with a Facet. The patch removes the dead code forgotten in the refactoring that introduced the concept of facets.
 - Modify the code loading package.json to ignore files in subdirectories, because we don't support workspaces with multiple projects.

See #181

/to @ritch please review


ritch:
LGTM

bajtos:
Released in `loopback-workspace@3.6.5`

PR #184, opened 2015-01-20
Add .strong-pm to .gitignore

sam-github:
Fix https://groups.google.com/d/topic/loopbackjs/T-W900RZ86I/discussion

I believe this to be the right place to make this change, though I'm not sure if the bug with .gitignore not being copied when a loopback app is scaffolded has been fixed yet.


bajtos:
LGTM.

PR #186, opened 2015-02-06
Rename .gitignore to gitignore

superkhau:
This is a workaround for NPM. When publishing NPM automatically renames
`.gitignore` files to `.npmignore`. Naming the file `gitignore` allow the file
to remain intact during publishing and we instead rename it back to
`.gitignore` as part of the scaffolding process instead.


bajtos:
1) This is not just a plain rename, you have modified the content of the file two. What's the purpose of that?

2) While the generator is the prime user of loopback-workspace, it is not the only possible user. After you have renamed the file, the projects templated directly via loopback-workspace will get `gitignore` instead of `.gitignore`.

Unfortunately there is no test for that yet, please add a new one to `test/workspace.js`.

3) The fact that `.gitignore` is stored as `gitignore` is an internal implementation detail of loopback-workspace, consumers of this module should not be required to deal with this problem.

A possible solution: `.gitignore` from `templates/components/api-server/template/.gitignore` to `templates/components/api-server/gitignore` and then add an extra call of `copyRecursive()` with appropriate parameters to perform the rename.

superkhau:
>1) This is not just a plain rename, you have modified the content of the file two. What's the purpose of that?

[ ] I've split the modifications into separate commits with individual commits to address these concerns.

>2) While the generator is the prime user of loopback-workspace, it is not the only possible user. After you have renamed the file, the projects templated directly via loopback-workspace will get gitignore instead of .gitignore.

Unfortunately there is no test for that yet, please add a new one to test/workspace.js.

[ ] Add test to verify projects templated directly are renamed to .gitignore properly

3) The fact that .gitignore is stored as gitignore is an internal implementation detail of loopback-workspace, consumers of this module should not be required to deal with this problem.

A possible solution: .gitignore from templates/components/api-server/template/.gitignore to templates/components/api-server/gitignore and then add an extra call of copyRecursive() with appropriate parameters to perform the rename.

[ ] I think you're missing a word here "...gitignore from templates...", move or copy?

bajtos:
> Move or copy? Need clarification.

Move `templates/components/api-server/template/.gitignore` to `templates/components/api-server/gitignore`. There should be always a single source of truth, in this case a single template for the gitignore file.

The purpose of this move is to exclude `.gitignore` from the current list of copied files. Once that's done, you can implement a custom copy/rename for gitignore.

superkhau:
@bajtos review again pls

bajtos:
Looks better. Please remove `templates/components/api-server/template/.gitignore` since it's no longer used.

bajtos:
@slnode test please

superkhau:
@bajtos Review again please. The test passes, but I notice `yo loopback` doesn't generate the `.gitignore` properly anymore. I'm not sure why yet thought, need to investigate what we need to change over there.

superkhau:
After more digging, I notice if I comment out https://github.com/strongloop/generator-loopback/blob/master/app/index.js#L45-L56 then everything copies over properly. Why do we bother overriding Workspace.copyRecursive with Yeoman this.directory?

superkhau:
OK, the problem is how Yeoman copies files (see http://yeoman.github.io/generator/actions.js.html#line325). Basically files `expandFiles` tries to copy nothing since we replace `ncp` with the native Yeoman `this.directory`. We have a few possible solutions:

1. Do not use the native `this.directory` (do not replace Workspace.copyRecursive and use the default `ncp` implementation)
  - *Pros*
    - Simple fix, just delete the `injectWorkspaceCopyRecursive` task
  - *Cons*
    - Not using native copy mechanisms
2. Add a task specifically to copy/rename `gitignore` to `.gitignore` in the destination dir
  - *Pros*
    - Use native copy mechanisms (`this.copy` or something)
  - *Cons*
    - More complicated to implement than deleting a few lines

@bajtos The issue above is for users who decide to replace copyRecursive with their own implementation that doesn't copy single files in the same way (ie. `ncp` vs Yeoman `this.directory`). Please let me know your thoughts and how to proceed.

bajtos:
loopback-generator is intentionally using native copy mechanism, as it provides simple conflict detection and resolution mechanism. I am very reluctant to let it go.

I am proposing to modify generator-loopback to run `fs.statSync` to check if the source is a file or a directory. When it is a file, run `this.copy`, otherwise run `this.directory`. Alternatively, you can change our generator to run `this.expandFiles` and `this.copy` instead of `this.directory`, and handle the situation when `this.expandFiles` does not return anything for `.gitignore`.

superkhau:
@bajtos Can I merge this PR then? I will do the other changes in `generator-loopback` itself. We can track that issue directly at https://github.com/strongloop/generator-loopback/issues/86

bajtos:
@slnode test please

bajtos:
Please rebase on top of https://github.com/strongloop/loopback-workspace/pull/189 and make sure the tests are passing.

Other than that, the patch LGTM.

superkhau:
@slnode test please

superkhau:
Link https://github.com/strongloop/strong-arc/issues/685

superkhau:
Link https://github.com/strongloop/strong-arc/issues/685

bajtos:
@superkhau I believe the correct keyword for waffle is "connect", i.e. Connect strongloop/strong-arc#685

superkhau:
@bajtos It seems like its every word except "connect", which is interesting. See https://github.com/waffleio/waffle.io/wiki/FAQs#prs-connect-keywords

tldr;
“Connect to”, “Connects to”, “Connected to”, "Connects"

bajtos:
Good to know!

PR #190, opened 2015-02-24
Upgrade ncp and strong-cached-install for io.js

bajtos:
Upgrade ncp to ^2.0 and strong-cached-install to ^2.0 in order to get support for io.js.

This change drops support for Node v0.8.

/cc @rmg @ritch

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "loopback-datasource-juggler": "^2.6.1",
     "method-override": "^2.1.1",
     "morgan": "^1.2.0",
-    "ncp": "^1.0.0",
+    "ncp": "^2.0.0",
     "serve-favicon": "^2.0.1",
     "strong-wait-till-listening": "^1.0.0",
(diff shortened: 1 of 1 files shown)


bajtos:
There is one thing I am not sure about. Should we release this as a major version, because we are dropping compatibility with Node v0.8? On the other hand, AFAIK, we never explicitly supported Node v0.8, in which case this can be considered a patch version. Thoughts?

rmg:
We don't test on 0.8, so as far as I know we don't support it.

bajtos:
> We don't test on 0.8, so as far as I know we don't support it.

Cool. I'll release this as a minor version as it can be considered as a new feature (adding support for new platforms) plus it will allow us to branch on the previous minor version in the rare case that somebody asks for 0.8 support.

bajtos:
Released as `loopback-workspace@3.8.0`.

PR #174, opened 2014-12-04
added cors middleware

claylo:
Enables CORS across the entire loopback application, using reasonable defaults. 

See https://github.com/strongloop/loopback/issues/575

PR #192, opened 2015-03-10
Setup a single top-level CORS middleware

bajtos:
This is a cleaned version of @claylo's pull request #174 which enables CORS across the entire loopback application, using reasonable defaults.

Close https://github.com/strongloop/loopback/issues/575

/to @raymondfeng or @ritch please review

### DIFF
--- a/templates/components/api-server/component.js
+++ b/templates/components/api-server/component.js
@@ -12,6 +12,7 @@ template.package = {
   },
   "dependencies": {
     "compression": "^1.0.3",
+    "cors": "^2.5.2",
     "errorhandler": "^1.1.1",
     "loopback": "^2.14.0",
     "loopback-boot": "^2.6.5",
@@ -67,10 +68,7 @@ template.server = {
           extended: true,
           limit: '100kb'
         },
-        cors: {
-          origin: true,
-          credentials: true
-        },
+        cors: false,
         errorHandler: {
           disableStackTrace: false
         }
--- a/templates/components/api-server/template/server/middleware.json
+++ b/templates/components/api-server/template/server/middleware.json
@@ -3,7 +3,14 @@
     "loopback#favicon": {}
   },
   "initial": {
-    "compression": {}
+    "compression": {},
+    "cors": {
+      "params": {
+        "origin": true,
+        "credentials": true,
+        "maxAge": 86400
+      }
+    }
   },
   "session": {
   },
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -80,6 +80,14 @@ describe('end-to-end', function() {
         .expect(200, done);
     });
 
+    it('provides CORS headers for all URLs', function(done) {
+      request(app).get('/')
+        .set('X-Requested-By', 'XMLHttpRequest')
+        .set('Origin', 'http://example.com')
+        .expect('Access-Control-Allow-Origin',  'http://example.com')
+        .expect(200, done);
+    });
+
     it('can create and login a user', function(done) {
       var credentials = { email: 'test@example.com', password: 'pass' };
       var userId, tokenId;


templates/components/api-server/component.js
```diff
@@ -67,10 +68,7 @@ template.server = {
           extended: true,
           limit: '100kb'
         },
-        cors: {
-          origin: true,
-          credentials: true
-        },
+        cors: false,
```
raymondfeng: Is it to disable model level `cors` as we now have a global one?

templates/components/api-server/template/server/middleware.json
```diff
@@ -3,7 +3,17 @@
...
   "initial": {
-    "compression": {}
+    "compression": {},
+    "cors": {
+      "params": {
+        "origin": true,
+        "credentials": true,
+        "exposedHeaders": true,
+        "methods": "GET,POST,PUT,PATCH,DELETE,OPTIONS",
+        "allowedHeaders": "Content-Type,Authorization",
```
raymondfeng: Is it too restrictive? What about leaving it as default:
```
If not specified, defaults to reflecting the headers specified in the request's Access-Control-Request-Headers header.
```

templates/components/api-server/template/server/middleware.json
```diff
@@ -3,7 +3,17 @@
...
   },
   "initial": {
-    "compression": {}
+    "compression": {},
+    "cors": {
+      "params": {
+        "origin": true,
+        "credentials": true,
+        "exposedHeaders": true,
+        "methods": "GET,POST,PUT,PATCH,DELETE,OPTIONS",
```
raymondfeng: What about `HEAD`?

templates/components/api-server/template/server/middleware.json
```diff
@@ -3,7 +3,17 @@
...
+        "allowedHeaders": "Content-Type,Authorization",
+        "maxAge": 180
```
raymondfeng: Maybe we should set `maxAge` to a higher number as we don't change `CORS` policies dynamically.

templates/components/api-server/component.js
bajtos: Yes, see https://github.com/strongloop/strong-remoting/pull/165

bajtos:
@claylo Could you please take a look at @raymondfeng's comments above regarding the settings you are proposing and explain why they are better than sticking with module's defaults?

Otherwise I am inclining towards using defaults as much as possible and reducing the scaffolded config to the following:

```json
{
  "origin": true,
  "credentials": true
}
```

Thoughts?

claylo:
What I chose as the default config in my PR was what seemed like reasonable defaults at the time. Now that I'm looking back a few months later, I can't think of good arguments for keeping it that way.

Fully implemented CORS support would modify the CORS options on a per-endpoint basis, so that pre-flight OPTIONS requests to a particular endpoint actually returns the HTTP methods the endpoint supports given the authentication credentials provided. :smile: 

Since that's a bigger project than this, I think @bajtos suggestion to keep the scaffolded config to a minimum is a good idea. It's great that it's supported server-wide with this change, and I think further config can be left as an exercise for each application's implementation.

Thanks, guys!

raymondfeng:
@bajtos Probably we should add `maxAge` to avoid aggressive preflight checks. See http://stackoverflow.com/questions/23543719/cors-access-control-max-age-is-ignored.

bajtos:
@raymondfeng what value of `maxAge` do you propose? Supposedly browsers cap this value at different values (Chrome - 10 mins, Safari - 5 mins, Firefox - 24 hours).

raymondfeng:
Set it to the highest value, such as `24 hours`.

bajtos:
@raymondfeng fixed, LGTY now?

raymondfeng:
LGTM

PR #194, opened 2015-04-17
Enable "validateUpsert" for newly created models

bajtos:
Connect to strongloop/loopback-datasource-juggler#262

/to @raymondfeng or @ritch please review

### DIFF
--- a/models.json
+++ b/models.json
@@ -131,7 +131,9 @@
       "idInjection": { "type": "boolean", "default": true },
       "scopes": "object",
       "indexes": "object",
-      "options": "object"
+      "options": { "type": "object", "default": {
+        "validateUpsert": true
+      } }
     },
     "public": true,
     "dataSource": "db",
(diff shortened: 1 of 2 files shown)


raymondfeng:
The patch LGTM. But in general, we need to have a way to roll out `semi-breaking` changes so that we can make them default in next major release.  Introducing a flag for each thing is annoying.

bajtos:
@slnode test pleaes

bajtos:
@slnode test please

bajtos:
> But in general, we need to have a way to roll out semi-breaking changes so that we can make them default in next major release. 

At the moment, I am keeping a GH issue with a list of things to remove: https://github.com/strongloop/loopback/issues/1224

> Introducing a flag for each thing is annoying.

I agree it's a bit annoying, but I don't have any better solution. Do you have one?

PR #198, opened 2015-05-07
api-server: configure default mixin sources

bajtos:
As a follow-up for https://github.com/strongloop/loopback-boot/pull/131, this patch modifies the scaffolded `model-config.json` to include lookup paths for mixins.

 - loopback/common/mixins
 - loopback/server/mixins
 - ../common/mixins
 - ./mixins

Connect to strongloop/loopback-boot#79

/to @ritch please review
/cc @PradnyaBaviskar

### DIFF
--- a/templates/components/api-server/component.js
+++ b/templates/components/api-server/component.js
@@ -44,6 +44,12 @@ template.server = {
         'loopback/server/models',
         '../common/models',
         './models',
+      ],
+      mixins: [
+        'loopback/common/mixins',
+        'loopback/server/mixins',
+        '../common/mixins',
+        './mixins',
       ]
     }
   },
--- a/test/facet.js
+++ b/test/facet.js
@@ -50,6 +50,12 @@ describe('Facet', function () {
           'loopback/server/models',
           '../common/models',
           './models'
+        ],
+        mixins: [
+          'loopback/common/mixins',
+          'loopback/server/mixins',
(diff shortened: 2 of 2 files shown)


templates/components/api-server/component.js
```diff
@@ -44,6 +44,12 @@ template.server = {
         'loopback/server/models',
         '../common/models',
         './models',
+      ],
+      mixins: [
```
bajtos: I am pondering whether "mixins" is a good key name, perhaps "mixinSources" would be a better name? (See also "sources" key used for model sources above.)

bajtos:
@ritch ping

templates/components/api-server/component.js
ritch: It should be consistent with what we have for models IMO

ritch:
LGTM

templates/components/api-server/component.js
bajtos: > It should be consistent with what we have for models IMO

:+1: for consistency

I am not sure how to apply that rule in this particular case. What property name would you find more consistent?

This is the current prologue of the scaffolded `model-config.json` file:

```json
{
  "_meta": {
    "sources": ["loopback/common/models", "etc."],
    "mixins": ["loopback/common/mixins", "etc."]
}
```

PR #199, opened 2015-05-12
api-server template: add strong-express-metrics

bajtos:
Modify the template to configure strong-express-metrics for all scaffolded applications (projects).

Close strongloop-internal/scrum-loopback#176

/to @raymondfeng @ritch please review


raymondfeng:
LGTM

PR #201, opened 2015-05-27
Revert "api-server template: add strong-express-metrics"

bajtos:
This reverts commit 6dc86830f5690bc247fbd6f3cbe9215479a29c80.

The module strong-express-metrics has changed the license to StrongLoop only, therefore it's better to not include it in the scaffolded applications by default.

/cc @raymondfeng @altsang @ritch 

I am going to land this without review as it's just `git revert {sha}`.

I am not sure whether we have to release this as a major version, I would prefer to make this a minor release instead. Thoughts?


ritch:
Minor is fine... LGTM

bajtos:
FWIW, the CI builds are failing on a timeout unrelated to this change. I'll merge.

PR #202, opened 2015-05-28
Fix ModelAccessControl.create() and failing tests

bajtos:
loopback-datasource-juggler has changed the signature of the `.create()` method by adding a new argument in the middle (see https://github.com/strongloop/loopback-datasource-juggler/pull/435). This caused a problem in our custom version of ModelAccessControl.create() that did not support three arguments.

/to @raymondfeng please review


raymondfeng:
LGTM

PR #203, opened 2015-06-02
Add middleware config model

raymondfeng:
/to @ritch 

Connect to strongloop/loopback#1428

### DIFF
--- a/models.json
+++ b/models.json
@@ -381,5 +381,36 @@
         }
       }
     }
+  },
+  "Middleware": {
+    "properties": {
+      "id": {"type": "string", "id": true, "json": false},
+      "name": {"type": "string", "json": false},
+      "phase": {"type": "string", "json": false},
+      "subPhase": {"type": "string", "json": false},
+      "isPhasePlaceHolder": {"type": "boolean", "json": false},
+      "nextPhase": {"type": "string", "json": false},
+      "order": {"type": "number", "json": false},
+      "paths": {"type": ["string"]},
+      "params": {"type": "object"},
+      "facetName": {"type": "string", "required": true, "json": false}
+    },
+    "public": true,
+    "dataSource": "db",
+    "options": {
+      "base": "Definition",
+      "defaultConfigFile": "middleware.json",
+      "configFiles": [
+        "middleware.json",
+        "middleware.*.json"
+      ],
+      "relations": {
+        "facet": {
+          "type": "belongsTo",
+          "model": "Facet",
+          "foreignKey": "facetName"
+        }
+      }
+    }
   }
 }
--- a/models/facet.js
+++ b/models/facet.js
@@ -5,6 +5,7 @@ var app = require('../app');
 
 var ModelDefinition = app.models.ModelDefinition;
 var DataSourceDefinition = app.models.DataSourceDefinition;
+var Middleware = app.models.Middleware;
 var ModelConfig = app.models.ModelConfig;
 var ConfigFile = app.models.ConfigFile;
 var FacetSetting = app.models.FacetSetting;
@@ -40,6 +41,7 @@ Facet.loadIntoCache = function(cache, facetName, allConfigFiles, cb) {
   var facetConfig = ConfigFile.getFileByBase(configFiles, 'config');
   var modelConfigs = ConfigFile.getFileByBase(configFiles, 'model-config');
   var dataSources = ConfigFile.getFileByBase(configFiles, 'datasources');
+  var middlewares = ConfigFile.getFileByBase(configFiles, 'middleware');
   var modelDefinitionFiles = ConfigFile.getModelDefFiles(configFiles, facetName);
   var steps = [];
 
@@ -132,7 +134,15 @@ Facet.loadIntoCache = function(cache, facetName, allConfigFiles, cb) {
         debug('loading [%s] dataSource into cache', dataSourceName);
         DataSourceDefinition.addToCache(cache, def);
       });
-      
+      cb();
+    });
+  }
+
+  if(middlewares) {
+    steps.push(function(cb) {
+      middlewares.load(cb);
+    }, function(cb) {
+      Middleware.deserialize(cache, facetName, middlewares);
       cb();
     });
   }
@@ -185,22 +195,27 @@ Facet.saveToFs = function(cache, facetData, cb) {
   }
 
   if (hasApp) {
-    var dataSoureConfig = {};
+    var dataSourceConfig = {};
     var dataSourcePath = path.join(facetName, 'datasources.json');
     var cachedDataSources = DataSourceDefinition.allFromCache(cache);
 
     cachedDataSources.forEach(function(dataSourceDef) {
       if(dataSourceDef.facetName === facetName) {
         dataSourcePath = DataSourceDefinition.getPath(facetName, dataSourceDef);
-        dataSoureConfig[dataSourceDef.name] =
+        dataSourceConfig[dataSourceDef.name] =
           DataSourceDefinition.getConfigFromData(dataSourceDef);
       }
     });
 
     addFileToSave(new ConfigFile({
       path: dataSourcePath,
-      data: dataSoureConfig
+      data: dataSourceConfig
     }));
+
+    var middlewareFile = Middleware.serialize(cache, facetName);
+    if (middlewareFile) {
+      addFileToSave(middlewareFile);
+    }
 
     var cachedModelConfigs = ModelConfig.allFromCache(cache);
     var modelConfigPath = path.join(facetName, ModelConfig.settings.defaultConfigFile);
--- a/models/middleware.js
+++ b/models/middleware.js
@@ -1,0 +1,296 @@
+var app = require('../app');
+var loopback = require('loopback');
+var debug = require('debug')('workspace:middleware');
+var ConfigFile = app.models.ConfigFile;
+var path = require('path');
+var stableSortInPlace = require('stable').inplace;
+
+/**
+ * Defines a `Middleware` configuration.
+ * @class Middleware
+ * @inherits Definition
+ */
+
+var Middleware = app.models.Middleware;
+
+/**
+ * - `facetName` is required and must refer to an existing facet
+ *
+ * @header Property Validation
+ */
+Middleware.validatesPresenceOf('facetName');
+
+function getPhase(name) {
+  if (name == null) {
+    return name;
+  }
+  var parts = name.split(':');
+  return parts[0];
+}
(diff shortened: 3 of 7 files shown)


models/facet.js
```diff
@@ -132,7 +134,15 @@ Facet.loadIntoCache = function(cache, facetName, allConfigFiles, cb) {
...
-      
+      cb();
+    });
+  }
+
+  if(middlewares) {
+    steps.push(function(cb) {
+      middlewares.load(cb);
+    }, function(cb) {
+      Middleware.deserialize(cache, facetName, middlewares);
```
bajtos: IIRC, other entities use different method names instead of serialize/deserialize, it would be nice to preserve consistency in naming.

test/middleware.js
```diff
@@ -0,0 +1,139 @@
...
+      });
+    });
+    it('should be persist multiple to the config file', function(done) {
+      var defs = Object.keys(this.configFile.data).sort();
+      expect(defs).to.eql(['auth', 'routes'].sort());
+      done();
+    });
+
+    it('should not contain workspace-private properties', function(done) {
+      // This test is reproducing an issue discovered in generator-loopback
```
bajtos: Is this comment true? Looks like a result of copy and paste to me.

models/facet.js
raymondfeng: Other entities have the logic embedded in facet.js, not as separate functions on the entity model itself.

models/middleware.js
```diff
@@ -0,0 +1,89 @@
...
+      def = defs[d];
+      def.configFile = configFile.path;
+      def.phase = phase;
+      def.facetName = facetName;
+      def.name = d;
+      debug('loading [%s] middleware into cache', def.name);
+      Middleware.addToCache(cache, def);
+    }
+    if (!def) {
+      // Keep the phase information by adding an empty middleware config
```
bajtos: AFAICT, there is not unit-test to verify how this special entry is handled. Could you please add one?

models/middleware.js
```diff
@@ -0,0 +1,89 @@
...
+      def = {
+        configFile: configFile.path,
+        phase: phase,
+        facetName: facetName,
+        name: '*' + phase // Set the name to be unique
```
bajtos: Is this unique enough? Can we use an internal property (a flag) instead?

models/facet.js
bajtos: I looked up the methods I was referring to: `ModelDefinition.getConfigFromCache`, `WorkspaceEntity.getConfigFromData`, `WorkspaceEntity.getDataFromConfig`. If I read the code in this patch correctly, your new serialize/deserialize methods call these `get*From*` method under the hood :+1: 

I agree it's a good idea to keep this logic out of facet.js.

Consider this comment as resolved then.

bajtos:
Could you please add a test to verify that the order of phases is preserved after a load/save cycle?

Also what is the API for adding a new (empty) middleware phase? Could you please write a test for that too? *As a workspace API user, I'd like to define (insert) a new middleware phase at a given position.*

ritch:
My biggest concern is breaking existing apps that use middleware. Writing a test that ensures order (according to lb runtime) is preserved for an app would ensure that won't happen.

bajtos:
Few more scenarios (edge cases?) to consider:
 - What is the API call to reorder middleware phases? (Think of Arc GUI implementing middleware & phase editor.)
 - What happens when the last middleware from a phase is removed, either by deleting the middleware or by moving it to a different phase? IMO, the empty middleware phase should be preserved. Consider the following scenario: remove the default error handler, add a custom one.

bajtos:
One more thing to consider: At the moment, project configuration that has a representation in Workspace API (Model, DataSource), is scaffolded via [component.js](https://github.com/strongloop/loopback-workspace/blob/05442a8754123a7fc77991f3c7a94b9d87470c26/templates/components/api-server/component.js) file. File templates that are copied as-is are used only for configuration that cannot be represented via Workspace API.

So far, there was no API for working with middleware, therefore the default middleware config is specified in a file [server/workspace.json](https://github.com/strongloop/loopback-workspace/blob/05442a8754123a7fc77991f3c7a94b9d87470c26/templates/components/api-server/template/server/middleware.json).

Now that we have API for middleware too, it is probably better to refactor `server/middleware.json` and move the template data to `component.js`, in order to preserve consistency.

raymondfeng:
@bajtos Added the ability to keep the order of entries and related tests. PTAL.

bajtos:
I think the proposed API will be rather difficult to use correctly (e.g. what is the API call to reorder middleware phases?), but if this is what works for your particular use case, then I am ok with it. I'll let @ritch to make a detailed review.

raymondfeng:
We can refine the (not end-user facing) apis over time based on the use cases. Maybe I'll start to add `slc loopback:middleware` generator to exercise the apis.

Ideally, we should have the MiddlewarePhase and Middleware models. The Middleware belongsTo MiddlewarePhase. The MiddlewarePhase defines the phases and their ordering.

bajtos:
> Maybe I'll start to add slc loopback:middleware generator to exercise the apis.

That would be great, as it would prove whether my concerns were relevant or not. I would personally like to see a consumer of this API implemented before we make a new release of loopback-workspace to npmjs.org, so that we don't have to bump up the major version if we find that this new API need breaking changes.

> Ideally, we should have the MiddlewarePhase and Middleware models. The Middleware belongsTo MiddlewarePhase. The MiddlewarePhase defines the phases and their ordering.

Sounds good to me :+1:

bajtos:
@slnode test please

bajtos:
@raymondfeng please consider extending the method `addPhase` with a check that the new phase is not inserted between sub-phases, i.e. a request to insert "audit" before "auth:after" should be rejected if there is "auth" or "auth:before". See https://github.com/strongloop/generator-loopback/pull/104#issuecomment-111213797 for more details.

raymondfeng:
@bajtos The latest version supports sub-phases. PTAL.

models.json
```diff
@@ -381,5 +381,69 @@
         }
       }
     }
+  },
+  "Middleware": {
+    "properties": {
+      "id": {"type": "string", "id": true, "json": false},
+      "name": {"type": "string", "json": false},
+      "phase": {"type": "string", "json": false},
```
bajtos: I think that `MiddlewarePhase` relation "hasMany phases" will add a property `Middleware.phaseId`. Is it a good idea to have both `phase` and `phaseId`? Should the model ensure that these two properties are always consistent, i.e. prevent a situation where `phase` points to a different entity than `phaseId`?

models/middleware.js
```diff
@@ -0,0 +1,296 @@
...
+ * @param {Function} cb Callback function
+ */
+Middleware.getPhases = function(cb) {
+  // Load all entries from cache and sort them by order
+  Middleware.find(function(err, instances) {
+    if (err) return cb(err);
+    var phases = sortMiddleware(instances).phases;
+    cb(null, phases);
+  });
+}
```
bajtos: I think this method belongs to `MiddlewarePhase`, what do you think?

models/middleware.js
```diff
@@ -0,0 +1,296 @@
...
+
+var ORDER_BUFFER = 1024;
+
+/**
+ * Add a middleware phase to the cache in a slot after the `nextPhase`
+ * @param {String} facetName
+ * @param {String} phase
+ * @param {String} nextPhase
+ */
+Middleware.addPhase = function(facetName, phase, nextPhase, cb) {
```
bajtos: Ditto, I think this method should be implemented by `MiddlewarePhase`. What are the benefits of keeping it here, in `Middleware`?

models.json
```diff
@@ -381,5 +381,69 @@
...
+      "relations": {
+        "facet": {
+          "type": "belongsTo",
+          "model": "Facet",
+          "foreignKey": "facetName"
+        }
+      }
+    }
+  },
+  "MiddlewarePhase": {
```
bajtos: It looks like `MiddlewarePhase` is defined in JSON, but it is never created nor referenced anywhere. It looks like a dead code to me. We should either remove this model, or else put it to use. Thoughts?

models/middleware.js
```diff
@@ -0,0 +1,296 @@
...
+ * @param {String} facetName Facet name
+ * @returns {ConfigFile}
+ */
+Middleware.serialize = function(cache, facetName) {
+  var middlewarePath = path.join(facetName, 'middleware.json');
+  var phases = loadFromCache(cache);
+  var middlewareConfig = {};
+
+  function addEntries(phase, subPhase, entries) {
+    if (Array.isArray(entries)) {
```
bajtos: Nitpick: use inverse condition + early return, it makes the code easier to understand and less nested.

```js
if (!Array.isArray(entries) return 0;
entries.forEach(function(m)  {
  // ...
});
return entries.length;
```

test/middleware.js
```diff
@@ -0,0 +1,212 @@
...
+          nextPhase: 'routes'
+        }, done);
+      }], done);
+    });
+
+    beforeEach(function(done) {
+      this.configFile.load(done);
+    });
+
+    it('should be able to create multiple', function(done) {
```
bajtos: "it should be able to create multiple" - multiple what? phases? middleware entries? something else?

test/middleware.js
```diff
@@ -0,0 +1,212 @@
...
+          if (err) done(err);
+          var middlewares = configFile.data;
+          expect(Object.keys(middlewares)).to.eql(
+            ['auth', 'routes:before', 'routes', 'myPhase']);
+          done();
+        });
+      });
+    });
+
+    it('should add phase after the first phase', function(done) {
```
bajtos: I think this adds the new phase at the end, i.e. the test should be called "...after the last phase".

test/middleware.js
```diff
@@ -0,0 +1,212 @@
...
+          if (err) done(err);
+          var middlewares = configFile.data;
+          expect(Object.keys(middlewares)).to.eql(
+            ['auth', 'routes:before', 'routes', 'myPhase', 'phase1']);
+          done();
+        });
+      });
+    });
+  });
+  it('validates `name` uniqueness within the facet only', function(done) {
```
bajtos: Please add an empty line between tests to visually delimit them.

bajtos:
@raymondfeng reviewed, see my comments above. The implementation looks good in general, I think the main question is what to do with `MiddlewarePhase` model. 

Please fix the failing CI build too.

models.json
raymondfeng: I'll get rid of MiddlePhase for now.

Issue #211, opened 2015-06-30
Implement the serialization/deserialization of Gateway Map/Pipeline/Policy models


raymondfeng:
We need to save/load policies.json by StrongLoop Arc and Gateway Runtime.

Issue #210, opened 2015-06-30
Add model/api for Gateway Map


raymondfeng:
A gateway map connects public endpoints to a pipeline. We need to have a way to specify endpoint patterns (VERB, URL) and references a pre-configured pipeline by name.

Issue #209, opened 2015-06-30
Add model/api for Gateway Policy


raymondfeng:
A policy implements the logic to intercept an api request and enforce certain QoSs.

Issue #208, opened 2015-06-30
Add model/api for Gateway Pipeline


raymondfeng:
Each pipeline references a list of configured policies.

PR #212, opened 2015-07-10
Add gateway policy related models

raymondfeng:
/to @ritch 
/cc @seanbrookes 

This PR adds models/apis for gateway related things based on the UX meetings

Connect to #208 
Connect to #209 
Connect to #210
Connect to #211


ritch:
@raymondfeng is your editor reformatting the JSON files? I see a ton of unrelated changes to the JSON structure. Tried `?w=1` but I still see all the changes to JSON formatting in the diff :(.

ritch:
@raymondfeng I don't think this belongs in the workspace. We are essentially adding a gateway only features into the generic LB workspace. My main concern is that when a breaking change is required for the gateway entities we will have to rev the major version of this module. This means fixes after that revision will not be pulled in by downstream modules (especially arc) until we upgrade them. This tightly couples the modules together. They would really share one global implicit version since when one changes the other must change.

Ideally these models would be implemented as a separate component depended on just by arc (and gateway specific modules). That way, breaking changes to just the gateway related entities can be made as often as required and the downstream can upgrade just that dependency.

I understand that this sounds like needless code structuring, but the workspace is already very complex and is depended on in many many places:

 - generator loopback
 - arc (composer)
 - explorer (the templating generates the code that adds the explorer)
 - all loopback applications (since you can add to a LB app using the workspace, applications depend on specific behavior of the workspace)
 - all loopback components

I think it will not be that hard to move this code into arc itself by registering it as a component on the workspace LB app (to avoid having to create yet another module). Then afterwards we can move it into a module if we need it.

raymondfeng:
@ritch I see your points but maybe the `gateway` term somewhat misleads you. Here is how I see the models:

1. We are adding a set of policy related models/apis. They are logically applicable to any LoopBack projects. 
2. We happened to pick up `GatewayMap` from the UX discussions but really that is mapping between resources (route-based) and policies.
3. The policies at the moment are bound to middleware but in the future they can be attached to other layers such as remoting, model hooks, or connector hooks.
4. It's similar for maps, we can potentially map policies to model/operation based resources.

PR #207, opened 2015-06-24
Update models.json to fix "methods" property to be considered a object

lius:
resolve https://github.com/strongloop/generator-loopback/issues/107

PR #218, opened 2015-08-04
Fix model methods to be represented as an object

bajtos:
This patch is a replacement for #207, which does not cleanly merge with the current master after the project layout was updated to the 2.0 structure.

Fix strongloop/generator-loopback#107

/to @ritch please review


raymondfeng:
LGTM

Issue #223, opened 2015-08-14
slc loopback related commands always  change excluded properties to properties


Neil-UWA:
`slc loopback` related commands will change excluded properties to properties.

If you exclude some properties, like email which inherits from base `User` model
```
{
  "properties": {
     "email": false
  }
}
```
when you run `slc loopback:XXX`, `slc` will change those `excluded properties` to model properties with wrong type definition,
```
{
  "properties": {
     "email": {
        "type": false
     }
  }
}
```

PR #224, opened 2015-08-14
Support flag to exclude inheritance of base properties

raymondfeng:
/to @ritch 

Fix for https://github.com/strongloop/loopback-workspace/issues/223

### DIFF
--- a/common/models/definition.js
+++ b/common/models/definition.js
@@ -108,16 +108,16 @@ module.exports = function(Definition) {
           var config = relatedData[embedId];
 
           if (relation.model === 'ModelProperty' && !(config && config.type)) {
-            if (typeof config === 'string' || typeof config === 'function') {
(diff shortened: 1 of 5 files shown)


raymondfeng:
It also fixes https://github.com/strongloop/loopback/issues/1325

raymondfeng:
Maybe a better way to exclude properties from being inherited from base is to add a separate attribute to the model definition, for example:

```js
{
  name: 'MyModelWithoutId',
  properties: {
  },
  base: 'PersistedModel',
  excludedBaseProperties: ['id']
}
```

raymondfeng:
Connect to #223

ritch:
Why do we even need this?

ritch:
Had an offline discussion with @raymondfeng.

LGTM

raymondfeng:
ok to test

PR #227, opened 2015-09-09
Allow array value for middleware definition

raymondfeng:
/to @bajtos 

This PR allows middleware.json for the following formats:
```json
"initial:before": {
    "strong-express-metrics": []
  },
"files": {
    "loopback#static": [
      {
        "paths": [
          "/"
        ],
        "params": "$!../client/public"
      },
      {
        "paths": [
          "/admin"
        ],
        "params": "$!../client/admin"
      }
    ]
}
```

### DIFF
--- a/common/models/middleware.js
+++ b/common/models/middleware.js
@@ -215,8 +215,21 @@ function ready(Middleware) {
             phaseDef = {};
             middlewareConfig[phaseName] = phaseDef;
           }
-          phaseDef[m.name] =
-            Middleware.getConfigFromData(m);
+          if (m.isMiddlewarePlaceHolder) {
+            phaseDef[m.name] = [];
+          } else {
+            var def = phaseDef[m.name];
+            if (def) {
+              // The name already has an entry, convert the value to array
+              if (!Array.isArray(def)) {
+                def = [def];
+              }
+              phaseDef[m.name] = def;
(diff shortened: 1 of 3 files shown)


raymondfeng:
Connect to strong-gateway#42

common/models/middleware.js
```diff
@@ -299,7 +324,11 @@ function ready(Middleware) {
     if (data.subPhase) {
       phase = phase + ':' + data.subPhase;
     }
-    return phase + '.' + data.name;
+    var index = '';
+    if (data.index) {
+      index = data.index.toString();
+    }
+    return phase + '.' + data.name + index;
```
bajtos: Please add a separator between `data.name` and `index` in order to prevent possible clashes.

Example:

```
phase: "files"
data.name: "serve-static"
index: 20

phase: "files"
data.name: "serve-static2"
index: 0
```

Suggested solution:

```js
return phase + '.' + data.name + '.' + index;
```

bajtos:
@raymondfeng the patch looks mostly good, see one comment above. I did not review details of the implementation, I'll trust you it works for you as needed.

Could you please add a link to loopback-boot pull request which implemented support for this new array format? Or was it supported from the beginning?

raymondfeng:
The array support is from the beginning, IIRC.

PR #236, opened 2015-11-26
Enhance security of scaffolded apps

gunjpan:
WIP: for **Apply "Best Practices for Express in Production" in project template**: https://github.com/strongloop-internal/scrum-loopback/issues/583 

Connect to strongloop-internal/scrum-loopback#583

### DIFF
--- a/templates/components/api-server/component.js
+++ b/templates/components/api-server/component.js
@@ -14,6 +14,7 @@ template.package = {
   "dependencies": {
     "compression": "^1.0.3",
     "cors": "^2.5.2",
+    "helmet": "^0.14.0",
     "loopback": "^2.22.0",
     "loopback-boot": "^2.6.5",
     "loopback-component-explorer": "^2.1.0",
--- a/templates/components/api-server/template/server/middleware.json
+++ b/templates/components/api-server/template/server/middleware.json
@@ -10,6 +10,24 @@
         "credentials": true,
         "maxAge": 86400
       }
+    },
+    "helmet#xssFilter": {},
+    "helmet#frameguard": {
+      "params": [ 
+        "deny"
+       ]
+    },
+    "helmet#hsts": {
+      "params": {
+        "maxAge": 0,
+        "includeSubdomains": true
+      }
+    },
+    "helmet#hidePoweredBy": {},
+    "helmet#ieNoOpen": {},
+    "helmet#noSniff": {},
+    "helmet#noCache": { 
+      "enabled": false
     }
   },
   "session": {
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -12,6 +12,7 @@ var models = workspace.models;
 var TestDataBuilder = require('loopback-testing').TestDataBuilder;
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
+var should = require('chai').should();
 
 var Workspace = workspace.models.Workspace;
 
@@ -96,6 +97,19 @@ describe('end-to-end', function() {
         .set('Origin', 'http://example.com')
         .expect('Access-Control-Allow-Origin',  'http://example.com')
         .expect(200, done);
+    });
+    
+    it('provides security headers for all URLs ', function(done){
+      request(app).get('/')
+       .expect('X-frame-options', 'DENY')
+       .expect('x-xss-protection', '1; mode=block')
+       .expect('x-content-type-options', 'nosniff')
+       .expect('x-download-options', 'noopen')
+       .expect(function(res){
+          var headers = res.headers;
+          headers.should.not.have.property('x-powered-by');
+       })
+       .expect(200, done);
     });
 
     it('can create and login a user', function(done) {


templates/components/api-server/component.js
```diff
@@ -14,6 +14,7 @@ template.package = {
   "dependencies": {
     "compression": "^1.0.3",
     "cors": "^2.5.2",
+    'helmet': "^0.14.0",
```
bajtos: Please use double-quotes here for consistency.

templates/components/api-server/template/server/server.js
```diff
@@ -1,7 +1,11 @@
 var loopback = require('loopback');
 var boot = require('loopback-boot');
+var helmet = require('helmet');
 
 var app = module.exports = loopback();
+//use helmet to improve app security: more details on Best practices for Express in Production:
+// https://strongloop.com/strongblog/best-practices-for-express-in-production-part-one-security/
+app.use(helmet());
```
bajtos: All middleware should be configured via `server/middleware.json`. To make it easier for users to tweak helmet config, the scaffolded `server/middleware.json` should include configuration mirroring helmet's defaults. That way users can just change a flag value, instead of having to look up all flag names in helmet docs.

gunjpan:
@bajtos : Please review. Moved Helmet config to middleware.json template and excluded CSP (test cases to follow soon). 
-  I've set some default(i.e. random) values for hpkp. We can instruct loopback-generator to modify hpkp settings. Your take on those hpkp default values, please. Thank you.

templates/components/api-server/template/server/middleware.json
```diff
@@ -10,6 +10,26 @@
         "credentials": true,
         "maxAge": 86400
       }
+    },
+    "helmet#xssFilter": {
+      "params": {
+        "setOnOldIE": true
```
bajtos: > On old versions of Internet Explorer, this creates a vulnerability (see here and here), and so the header is set to 0 to disable it. To force the header on all versions of IE, add the option:

I believe we should not enable this option by default, as it would create a security vulnerability.

templates/components/api-server/template/server/middleware.json
```diff
@@ -10,6 +10,26 @@
...
+      }
+    },
+    "helmet#frameguard": {
+      "params": "deny"
```
bajtos: I'd prefer to wrap the value in an array, to make it more clear how to pass more than one item.

```js
// default in the scaffolding
"params": ["deny"]
// an example of a custom config
"params": ["allow-from", "http://example.com"]
```

templates/components/api-server/template/server/middleware.json
```diff
@@ -10,6 +10,26 @@
...
+    },
+    "helmet#hsts": {
+      "params": {
+        "maxAge": 0,
+        "includeSubdomains": true
+      }
+    },
```
bajtos: Ha, I didn't realise "hsts" sends the header only for HTTPs requests. That's great, as it allows us to include the middleware by default :+1:

bajtos:
@gunjpan thank you for the pull request, the proposed changes look mostly good. Could you please rebase this patch on top of the current master?

Also could you please add automated tests to verify the implementation? I think they should go to [test/end-to-end.js](https://github.com/strongloop/loopback-workspace/blob/07b1cec1d7b773cf967a1c91c0022af6dfc80121/test/end-to-end.js) and simply verify that the app serves expected security headers.

test/end-to-end.js
```diff
@@ -97,6 +98,20 @@ describe('end-to-end', function() {
...
+       .expect('x-xss-protection', '1; mode=block')
+       .expect('x-content-type-options', 'nosniff')
+       .expect('x-download-options', 'noopen')
+       .expect(function(res){
+          var headers = res.headers;
+          console.log(headers);
+          headers.should.not.have.property('x-powered-by');
+       })
+       .expect(200, done);
+    });
```
gunjpan: @bajtos : Please review

test/end-to-end.js
```diff
@@ -97,6 +98,20 @@ describe('end-to-end', function() {
...
+    
+    it('provides security headers for all URLs ', function(done){
+      request(app).get('/')
+       .expect('X-frame-options', 'DENY')
+       .expect('x-xss-protection', '1; mode=block')
+       .expect('x-content-type-options', 'nosniff')
+       .expect('x-download-options', 'noopen')
+       .expect(function(res){
+          var headers = res.headers;
+          console.log(headers);
```
bajtos: Is this log here to work around insufficient error details printed by `headers.should.not.have.property`?

I am proposing to use `debug` instead.

```js
// at the top
var debug = require('debug')('test');

// here:
debug(headers);
```

test/end-to-end.js
```diff
@@ -731,7 +746,7 @@ function execNpm(args, options, cb) {
 
 function installSandboxPackages(cb) {
   this.timeout(300 * 1000);
-  install(SANDBOX, PKG_CACHE, ['dependencies', 'devDependencies'], cb);
+    install(SANDBOX, PKG_CACHE, ['dependencies', 'devDependencies'], cb);
```
bajtos: This looks like an unrelated whitespace change to me, please revert.

bajtos:
One minor comment, the patch LGTM otherwise.

test/end-to-end.js
gunjpan: @bajtos : Grrrr... I forgot to remove `console.log`, it was for my own reference only...
I believe, `headers.should.not.have.property` throws enough error details, I will remove the log part.

bajtos:
Landed, thank you!

PR #237, opened 2015-12-02
Fix: ModelMethod.http is an array

bajtos:
/to @ritch please review
/cc @jannyHou this affects your https://github.com/strongloop/generator-loopback/pull/124. For the first version, I think it's ok to support only one http endpoint in the generator.

### DIFF
--- a/common/models/model-method.json
+++ b/common/models/model-method.json
@@ -31,7 +31,7 @@
       "type": "string"
     },
     "http": {
-      "type": "object"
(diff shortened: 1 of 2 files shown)


jannyHou:
@bajtos Thank you for informing me about this, it's changed to array now(https://github.com/strongloop/generator-loopback/pull/124)

ritch:
LGTM

PR #239, opened 2015-12-09
Add post test script for security check.

gunjpan:
Enable post-test `nsp` check for known vulnerable
dependencies in sacffolded pacakage config.
Connect to strongloop-internal/scrum-loopback#583


bajtos:
Landed, thank you!

PR #254, opened 2016-02-22
Update dependencies

bajtos:
Update all dependencies to their latest (major) versions, with the exception of the following two dev dependencies:

 - `chai`, because they dropped function assertions in favour of property assertions
 - `grunt-docular`, because the module was always fragile and we don't use it much anyways


bajtos:
@slnode test please

PR #255, opened 2016-02-23
Add more templates

bajtos:
I this patch, I am cleaning up the directory structure in `templates` folder a bit, introducing template inheritance and new templates "empty-server" and "hello-world". The exiting template "api-server" is inheriting from "empty-server" to reuse as much template data as possible.

I am also adding `describeAvailableTemplates` and `template.description` property, this can be used to populate UI prompts with more meaningful texts.

/to @ritch please review, this patch is needed for "apic loopback template prompt".
/cc @kraman FYI


bajtos:
The next commit adds "displayName" and `Workspace.describeAvailableTemplates(cb)` that can be called to query this new metadata.

ritch:
I think some subtle things will need to change, but the overall impl LGTM. :+1:

PR #256, opened 2016-02-25
Package.json minor tweaks (version/private)

ritch:
@bajtos


bajtos:
@ritch please rebase your patch on top of the current master, #255 (and more specifically 32d0b49) changed the directory structure in templates:
> - Rename `templates/components` to `templates/projects`.
> - Rename `{name}/template` to `{name}/files`

What is the rationale for "Set private true for base package.json"?

bajtos:
Hmm, perhaps I am missing something here. AFAICT, your first commit modifies `templates/projects/empty-server/data.js`, which is the new (renamed) location. But the at the same time, you are adding a bunch of files in `templates/components/empty` and `templates/components/hello-world` - I believe these files won't be picked by the templating code. What is your intention here?

bajtos:
@ritch this pull request looks pretty abandoned and it's not entirely clear to me what was your intent, what and why you wanted to change. I am closing it, please open a new one based on the latest `master` codebase if this is still relevant.

PR #261, opened 2016-03-09
Add email-connector in `available-connectors.json`

Amir-61:
###### Notice:
However, I believe We don't have support to read an array or objects inside an object  in `settings` for different connectors in `generator-loopback`; please see [here](https://github.com/strongloop/generator-loopback/blob/a8a49cdda819d5378cefcc525a19d859f5901374/datasource/index.js#L111-L179). If I'm right we should support them? Also, if we decide to add that support I guess we need to clarify our approaches...

Thanks  a lot Miroslav :-)

Connect to: https://github.ibm.com/apimesh/scrum-platform-foundation/issues/352

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -27,6 +27,35 @@
       "discovery": false,
       "migration": false
     },
+    "settings":{
+      "transports": [{
+        "type": {
+          "type": "string"
+        },
+        "host": {
+          "type": "string"
+        },
+        "secure": {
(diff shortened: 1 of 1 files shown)


Amir-61:
@raymondfeng PTAL

Amir-61:
Closing in favour of : https://github.com/strongloop/loopback-workspace/pull/263

PR #263, opened 2016-03-10
Reorder the list of connectors and add settings for soap/rest/mail

raymondfeng:
The PR refines the list of connectors:
1. Reorder 
2. Add settings for soap/rest/mail connectors


chandadharap:
LGTM. (Eventually fro future, we should figure out a way to use the json to mark IBM supported and community ones in a way that UI can pick up).

PR #265, opened 2016-03-15
Remove model definition file when deleting by id

superkhau:


superkhau:
@raymondfeng Review pls.

superkhau:
Also, I'm not sure if fire and forget is a good solution, but its the quickest to impl ATM. I figure you might want me to do something like async.parallel the 2 deletes and send a fail callback if anyone fails. Let me know the preferred solution if this MVP one isn't good enough.

superkhau:
@slnode test please

common/models/model-definition.js
```diff
@@ -125,6 +125,10 @@ function ready(ModelDefinition) {
       removeById(id, function(err) {
         if (err) return cb(err);
 
+        // fire and forget since we want to remove the other files even if this
+        // one doesn't exist
+        fs.unlink(modelDef.getScriptPath());
```
raymondfeng: I think `fs.unlink()` takes a callback as the 2nd argument.

raymondfeng:
@superkhau You can delete things in parallel or series.

superkhau:
@raymondfeng Review again pls.

raymondfeng:
LGTM

PR #266, opened 2016-03-16
Fix regression due to change in delete API

jtary:
Fix regression due to change in the data format returned by deleteById.
Wrap the array returned by async.parallel into an object.

### DIFF
--- a/common/models/model-definition.js
+++ b/common/models/model-definition.js
@@ -136,7 +136,11 @@ function ready(ModelDefinition) {
         async.parallel([
           removeModelDef,
           removeModelDefJs
-        ], cb);
+        ], function(err, results) {
+          if (err) return cb(err);
+
+          cb(null, {result: results});
+        });
       });
     });
   }


common/models/model-definition.js
```diff
@@ -136,7 +136,9 @@ function ready(ModelDefinition) {
         async.parallel([
           removeModelDef,
           removeModelDefJs
-        ], cb);
+        ], function(results) {
```
raymondfeng: `async.parallel` returns an array and the order depends on which task finish 1st. I wonder if your change will break `strong-arc`.

@superkhau I think you should work with @jtary to make sure the callback returns the same result before your PR.

jtary: This used to return an object, but @superkhau change to use async.parallel changes it to an array which is what caused the regression.

superkhau: @raymondfeng Sounds good

superkhau:
@jtary Updated to handle error case, LGTY?

jtary:
LGTM

ritch:
@jtary @superkhau did the dependency job pass on this PR? I'm getting failures on generator-loopback... ?

superkhau:
@ritch No, the strong arc wasn't passing, but it was failing before this PR anyways.

PR #273, opened 2016-04-07
Update project to remove isStatic flag

0candy:
Connect to https://github.com/strongloop-internal/scrum-loopback/issues/813

Only for LB 3.0

### DIFF
--- a/common/models/definition.js
+++ b/common/models/definition.js
@@ -128,12 +128,13 @@ module.exports = function(Definition) {
               fileData.name, embedId, config);
           }
 
-          config = Entity.getDataFromConfig(config);
+          config = Entity.getDataFromConfig(config, embedId);
 
           // add extra properties for relations
           config[relation.foreignKey] = fk;
           config[relation.embed.key] = embedId;
           config.facetName = facetName;
+          
           debug('addRelatedToCache %s %j', relation.model, config);
           Entity.addToCache(cache, config);
         });
--- a/common/models/model-definition.js
+++ b/common/models/model-definition.js
@@ -75,10 +75,19 @@ function ready(ModelDefinition) {
     assert(relation.embed && relation.embed.as, 'embed requires "as"');
     switch (relation.embed.as) {
       case 'object':
-        assert(relation.embed.key, 'embed as object requires "key"');
+        assert(relation.embed.key || relation.embed.keyGetter,
+          'embed as object requires "key" or "keyGetter"');
         result = {};
         relatedData.forEach(function(related) {
-          var key = related[relation.embed.key];
+          var Definition = app.models[relation.model];
+          var key;
+          if (relation.embed.key) {
+            key = related[relation.embed.key];
+          }
+          var keyGetter = relation.embed.keyGetter;
+          if (keyGetter && typeof Definition[keyGetter] === 'function') {
+            key = Definition[keyGetter](related.name, related);
+          }
           result[key] = related;
         });
         cleanRelatedData(result, relation);
--- a/common/models/model-definition.json
+++ b/common/models/model-definition.json
@@ -96,7 +96,7 @@
     "methods": {
       "embed": {
         "as": "object",
-        "key": "name"
+        "keyGetter": "getJsonKey"
       },
       "type": "hasMany",
       "model": "ModelMethod",
--- a/common/models/model-method.js
+++ b/common/models/model-method.js
@@ -2,6 +2,7 @@
 // Node module: loopback-workspace
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
+var semver = require('semver');
 
 module.exports = function(ModelMethod) {
   /**
@@ -10,4 +11,49 @@ module.exports = function(ModelMethod) {
    * @class ModelMethod
    * @inherits WorkspaceEntity
    */
+
+  ModelMethod._shouldEncodeStaticFlagInName = function() {
+    var version = ModelMethod.app.models.Workspace.loopBackVersion;
+    return version != null ? !semver.gtr('3.0.0', version) : false;
+  };
+
+  ModelMethod.getJsonKey = function(name, data) {
+    if (!this._shouldEncodeStaticFlagInName()) {
+      return data.name;
+    }
+    var isStatic = data.isStatic;
+    if (isStatic !== undefined) {
+      var matchName = name.match(/^prototype\.(.*)$/);
+      if (!isStatic && (matchName === null || !matchName)) {
+        data.name = 'prototype.' + name;
+      }
+    }
+
+    return data.name;
+  };
+
+  ModelMethod.getConfigFromData = function(data) {
+    var config = ModelMethod.base.getConfigFromData.call(this, data);
+
+    if (this._shouldEncodeStaticFlagInName()) {
+      delete config.isStatic;
+    }
+    delete config.name;
+
+    return config;
+  };
+
+  ModelMethod.getDataFromConfig = function(config, name) {
+    var data = ModelMethod.base.getDataFromConfig.call(this, config);
+    data.name = name;
+
+    if (this._shouldEncodeStaticFlagInName()) {
+      var m = name.match(/^prototype\.(.*)$/);
+      var isStatic = !m;
+      data.name = isStatic ? name : m[1];
+      data.isStatic = isStatic;
+    }
+
+    return data;
+  };
 };
--- a/test/model-method.js
+++ b/test/model-method.js
@@ -101,3 +101,199 @@ describe('ModelMethod', function() {
       });
   });
 });
+
+describe('ModelMethod - Loopback 2.0', function() {
+  var userModel;
+
+  beforeEach(givenLB2Workspace);
+
+  beforeEach(function(done) {
+    ModelDefinition.create(
+      {
+        name: 'user',
+        facetName: 'server',
+      },
+      function(err, result) {
+        if (err) return done(err);
+        userModel = result;
+        done();
+      });
+  });
+
+  it('correctly adds static method with isStatic flag', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'testMethod',
+        isStatic: true,
+      },
+      function(err) {
+        if (err) return done(err);
+        userModel.methods(function(err, list) {
+          if (err) return done(err);
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.have.property('isStatic', true);
+
+          var cfg = new ConfigFile({ path: 'server/models/user.json' });
+          cfg.load(function(err) {
+            if (err) return done(err);
+            var methods = cfg.data.methods;
+            expect(methods).to.be.an('object');
+            expect(methods).to.have.property('testMethod');
+            expect(methods).to.not.have.property('prototype.testMethod');
+            expect(methods.testMethod).to.have.property('isStatic', true);
+            expect(methods.testMethod).to.not.have.property('id');
+            expect(methods.testMethod).to.not.have.property('facetName');
+            expect(methods.testMethod).to.not.have.property('name');
+            done();
+          });
+        });
+      });
+  });
+
+  it('correctly adds prototype method', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'testMethod',
+        isStatic: false,
+      },
+      function(err) {
+        if (err) return done(err);
+        userModel.methods(function(err, list) {
+          if (err) return done(err);
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.have.property('isStatic', false);
+
+          var cfg = new ConfigFile({ path: 'server/models/user.json' });
+          cfg.load(function(err) {
+            if (err) return done(err);
+            var methods = cfg.data.methods;
+            expect(methods).to.be.an('object');
+            expect(methods).to.have.property('testMethod');
+            expect(methods).to.not.have.property('prototype.testMethod');
+            expect(methods.testMethod).to.have.property('isStatic', false);
+            expect(methods.testMethod).to.not.have.property('id');
+            expect(methods.testMethod).to.not.have.property('facetName');
+            expect(methods.testMethod).to.not.have.property('name');
+            done();
+          });
+        });
+      });
+  });
+});
+
+describe('ModelMethod - Loopback 3.0', function() {
+  var userModel;
+
+  beforeEach(givenLB3Workspace);
+
+  beforeEach(function(done) {
+    ModelDefinition.create(
+      {
+        name: 'user',
+        facetName: 'server',
+      },
+      function(err, result) {
+        if (err) return done(err);
+        userModel = result;
+        done();
+      });
+  });
+
+  it('add static method without isStatic flag to method definition', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'testMethod',
+        isStatic: true,
+      },
+      function(err) {
+        if (err) return done(err);
+        userModel.methods(function(err, list) {
+          if (err) return done(err);
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.have.property('isStatic', true);
+
+          var cfg = new ConfigFile({ path: 'server/models/user.json' });
+          cfg.load(function(err) {
+            if (err) return done(err);
+            var methods = cfg.data.methods;
+            expect(methods).to.be.an('object');
+            expect(methods).to.have.property('testMethod');
+            expect(methods).to.not.have.property('prototype.testMethod');
+            expect(methods.testMethod).to.not.have.property('isStatic');
+            expect(methods.testMethod).to.not.have.property('id');
+            expect(methods.testMethod).to.not.have.property('facetName');
+            expect(methods.testMethod).to.not.have.property('name');
+            done();
+          });
+        });
+      });
+  });
+
+  it('add `prototype.` to method name if isStatic flag is false', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'testMethod',
+        isStatic: false,
+      },
+      function(err) {
+        if (err) return done(err);
+        userModel.methods(function(err, list) {
+          if (err) return done(err);
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.have.property('isStatic', false);
+
+          var cfg = new ConfigFile({ path: 'server/models/user.json' });
+          cfg.load(function(err) {
+            if (err) return done(err);
+            var methods = cfg.data.methods;
+            expect(methods).to.be.an('object');
+            expect(methods).to.have.property('prototype.testMethod');
+            expect(methods).to.not.have.property('testMethod');
+            expect(methods['prototype.testMethod']).to.not.have.property('isStatic');
+            expect(methods['prototype.testMethod']).to.not.have.property('id');
+            expect(methods['prototype.testMethod']).to.not.have.property('facetName');
+            expect(methods['prototype.testMethod']).to.not.have.property('name');
+            done();
+          });
+        });
+      });
+  });
+
+  it('loading JSON should have correct method name and isStatic flag', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'testMethod',
+        isStatic: false,
+        http: [
+          { verb: 'get', path: '/get' },
+          { verb: 'head', path: '/head' },
+        ],
+      },
+      function(err) {
+        if (err) return done(err);
+        userModel.methods(function(err, list) {
+          if (err) return done(err);
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.have.property('isStatic');
+
+          ModelMethod.find(function(err, methods) {
+            if (err) return done(err);
+            expect(methods[0]).to.have.property('name', 'testMethod');
+            expect(methods[0]).to.not.have.property('name', 'prototype.testMethod');
+            expect(methods[0]).to.have.property('isStatic', false);
+            done();
+          });
+        });
+      });
+  });
+});
--- a/test/support.js
+++ b/test/support.js
@@ -107,6 +107,36 @@ givenWorkspaceFromTemplate = function(template, options, cb) {
   });
 };
 
+givenWorkspaceWithCustomDependencies = function(templateName, deps, cb) {
+  resetWorkspace(function(err) {
+    if (err) return cb(err);
+    givenWorkspaceFromTemplate('empty-server', deps, function(err) {
+      if (err) return cb(err);
+      ConfigFile.loadFromPath('/package.json', function(err, pkgFile) {
+        if (err) return cb(err);
+        for (dep in deps) {
+          pkgFile.data.dependencies[dep] = deps[dep];
+        }
+
+        pkgFile.save(cb());
+      });
+    });
+  });
+};
+
+givenLB2Workspace = function(cb) {
+  givenWorkspaceWithCustomDependencies('empty-server', {
+    loopback: '^2.0.0',
+    'loopback-datasource-juggler': '^2.0.0',
+  }, cb);
+};
+
+givenLB3Workspace = function(cb) {
+  givenWorkspaceWithCustomDependencies('empty-server', {
+    loopback: '^3.0.0-alpha.1',
+  }, cb);
+};
+
 function findOfType(name, type) {
   assert(name);
   assert(type);
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "method-override": "^2.1.1",
     "morgan": "^1.2.0",
     "ncp": "^2.0.0",
+    "semver": "^5.1.0",
     "serve-favicon": "^2.0.1",
     "stable": "^0.1.5",
     "strong-wait-till-listening": "^1.0.0"
--- a/server/connector.js
+++ b/server/connector.js
@@ -199,6 +199,12 @@ connector._loadFromFile = function(cb) {
         f.load(function(err) {
           if (err) return next(err);
           PackageDefinition.addToCache(cache, f.data);
+          if (dir === '.') {
+            var loopBackVersion = f.data.dependencies['loopback'] ||
+              f.data.devDependencies['loopback'] ||
+              f.data.optionalDependencies['loopback'];
+            app.models.Workspace.loopBackVersion = loopBackVersion;
+          }
           next();
         });
       }, function(err) {


bajtos:
> Only for LB 3.0

There is no LB 3.0 version of loopback-workspace. We need the same version of loopback-workspace to support both LB 2.x and LB 3.x applications.

common/models/model-method.json
```diff
@@ -18,9 +18,6 @@
     "aliases": {
       "type": "array"
     },
-    "isStatic": {
-      "type": "boolean"
-    },
```
bajtos: I think we should keep this property around, so that consumers of loopback-workspace do not need to understand how exactly are remote methods described in JSON files. That way the API of loopback-workspace remains the same, which means less work for us too.

common/models/model-method.js
```diff
@@ -2,6 +2,8 @@
 // Node module: loopback-workspace
 // This file is licensed under the MIT License.
 // License text available at https://opensource.org/licenses/MIT
+var app = require('../../server/server');
```
bajtos: Models in "common" should not depend on things in "server", that goes against the direction of dependencies. You can access the app via `ModelMethod.app` after the model was configured (attached to a datasource & app).

common/models/model-method.js
```diff
@@ -10,4 +12,53 @@ module.exports = function(ModelMethod) {
...
+  });
 };
+
+
+function ready(ModelMethod) {
+  var debug = require('debug')('workspace:model-method');
+  var PackageDefinition = app.models.PackageDefinition;
+  var ModelDefinition = app.models.ModelDefinition;
+  
+  PackageDefinition.observe('after save', function getPackageDef(ctx, next) {
```
bajtos: I find this suspicious. What if `PackageDefinition` is never saved and only loaded from fs?

common/models/model-method.js
```diff
@@ -10,4 +12,53 @@ module.exports = function(ModelMethod) {
...
+    if (!pkg) {
+      debug('Cannot find Loopback version, defaulting to 2.x.');
+    } else {
+      var version = pkg.dependencies['loopback'];
+      var range = semver.Range(version);
+       // pkg.devDependencies['loopback'] = '1.x';
+      // pkg.optionalDependencies['loopback'] = '1.x';
+      debug('Loopback version: %s', version);
+    
+      if (semver.satisfies('3.0.0', range)) {
```
bajtos: 1) I think this will not work, what if the range is `^3.10.0`? The value "3.0.0" will not satisfy that range.

2) Please move this code to a method in `PackageDefinition`, e.g. `PackageDefinition.getLoopBackVersion()`.

common/models/model-method.js
```diff
@@ -10,4 +12,53 @@ module.exports = function(ModelMethod) {
...
+          var isStatic = data.isStatic;
+          if (isStatic !== undefined) {
+            var name = data.name;
+            var matchName = name.match(/^prototype\.(.*)$/);
+            if (!isStatic && (matchName === null || !matchName)) {
+              (ctx2.instance || ctx2.data).name = 'prototype.' + name;
+            }
+          }
+          // delete properties.isStatic;
+          delete ctx2.instance.__data.isStatic;
```
bajtos: 1) If you want to unset a property, you should call `ctx2.instance.unsetAttribute('isStatic')`.

2) I find this implementation rather hacky. If I am reading the code correctly, it will leave `ctx.instance` in a different state that no longer matches the API contract.

test/model-method.js
```diff
@@ -68,7 +67,6 @@ describe('ModelMethod', function() {
       {
         modelId: userModel.id,
         name: 'multiMethod',
-        isStatic: true,
```
bajtos: I find the changes in the existing test suite suspicious. Since we assume LB 2.0 by default, I would expect the current test cases should keep passing without any changes even with the new implementation.

test/model-method.js
```diff
@@ -101,3 +99,106 @@ describe('ModelMethod', function() {
...
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.not.have.property('isStatic');
+
+          var cfg = new ConfigFile({ path: 'server/models/user.json' });
+          cfg.load(function(err) {
+            if (err) return done(err);
+            var methods = cfg.data.methods;
+            expect(methods).to.be.an('object');
+            expect(methods).to.have.property('testMethod');
```
bajtos: Please add a check that the other name variant (`prototype.testMethod`) is not set.

```js
expect(methods).to.not.have.property('prototype.testMethod');
```

Just to be sure...

Please do the same in the other tests listed below too.

test/support.js
```diff
@@ -107,6 +107,13 @@ givenWorkspaceFromTemplate = function(template, options, cb) {
   });
 }
 
+givenLB3Workspace = function(cb) {
+  resetWorkspace(function(err) {
+    if(err) return cb(err);
+    givenWorkspaceFromTemplate('empty-server-3.0', cb);
```
bajtos: > templates/projects/empty-server-3.0/data.js

I think it is not a good idea to add a 3.0 template, because these templates are public and thus will be listed in `apic loopback` prompt.

I think it's better to use the existing template (`empty-server`) and manually change loopback version in the scaffolded package.json.

Around the time when we will be releasing LB 3.0, we will change empty-server to scaffold LB 3.0 apps. At that point, we need to change these tests so that existing remote-method tests depending on LB 2.0 keep using LB 2.0 projects. To make that transition easier, I am proposing to add another helper `givenLB2Workspace`. Both `givenLB*Workspace` helpers should create workspace from the same template and then explicitly change loopback version in `package.json`. With those helpers in place, `describe('ModelMethod')` should be renamed to `describe('ModelMethod - LoopBack 2.x')` and call `givenLB2Workspace`.

common/models/model-definition.js
```diff
@@ -76,10 +76,19 @@ function ready(ModelDefinition) {
     assert(relation.embed && relation.embed.as, 'embed requires "as"');
     switch(relation.embed.as) {
       case 'object':
-        assert(relation.embed.key, 'embed as object requires "key"');
+        assert(relation.embed.key || relation.embed.keyFn, 'embed as object requires "key" or "keyFn"');
         result = {};
         relatedData.forEach(function(related) {
-          var key = related[relation.embed.key];
+          var Entity = require('loopback').getModel(relation.model);
```
bajtos: Please use `app.models[relation.model]` to stay consistent with L65 above.

common/models/model-definition.js
```diff
@@ -76,10 +76,19 @@ function ready(ModelDefinition) {
...
+          var key;
+          if (relation.embed.key) {
+            key = related[relation.embed.key];
+          }
+          if (relation.embed.keyFn) {
+            if (typeof Entity[relation.embed.keyFn] === "function") {
```
bajtos: Please combine these two `if` statements into a single one:

```js
var keyFn = relation.embed.keyFn;
if (keyFn && typeof Entity[keyFn] === 'function') {
  key = Entity[keyFn](related);
}
```

common/models/model-definition.json
```diff
@@ -96,7 +96,8 @@
     "methods": {
       "embed": {
         "as": "object",
-        "key": "name"
+        "key": "name",
```
bajtos: I think `key` should not be set when `keyFn` is provided, should it?

common/models/model-method.js
```diff
@@ -10,4 +10,43 @@ module.exports = function(ModelMethod) {
...
+  }
+
+  ModelMethod.getConfigFromData = function(data) {
+    var version = ModelMethod.app.models.Workspace.loopBackVersion;
+    var properties = this.definition.properties;
+    var result = {};
+    var prop;
+
+    if (encodeStaticFlagInName(version)) {
+      delete data.isStatic;
```
bajtos: I think we need to add `prototype.` prefix to the name when `isStatic` is false.

bajtos: Ah, ignore my comment, the prefix is added by `getJsonKey`.

common/models/model-method.js
```diff
@@ -10,4 +10,43 @@ module.exports = function(ModelMethod) {
...
+    }
+    
+    for (prop in properties) {
+      if (properties[prop].json === false) continue;
+      result[properties[prop].json || prop] = data[prop];
+    }
```
bajtos: Please call parent (super) method instead of duplicating this code. Example:

```js
ModelMethod.getConfigFromData = function(data) {
  var config = ModelMethod.base.getConfigFromData(data);
  if (encodeStaticFlagInName(version)) {
    delete config.isStatic;
  }
  return config;
}
```

common/models/model-method.js
```diff
@@ -10,4 +10,43 @@ module.exports = function(ModelMethod) {
...
+
+    return result;
+  };
 };
+
+function encodeStaticFlagInName(version) {
```
bajtos: In order to avoid repetition of `ModelMethod.app.models.Workspace.loopBackVersion`, I am proposing to change this function into a private model method.

```js
ModelMethod._shouldEncodeStaticFlagInName = function() {
  var version = ModelMethod.app.models.Workspace.loopBackVersion; 
  // ...
};
```

Thoughts?

common/models/model-method.js
```diff
@@ -10,4 +10,43 @@ module.exports = function(ModelMethod) {
...
+  var semver = require('semver');
```
bajtos: The usual convention is to put all `require` calls at the top of the file. Is there any particular reason for doing it differently here?

common/models/package-definition.js
```diff
@@ -30,4 +30,15 @@ function ready(PackageDefinition) {
     configFile.save(cb);
   };
 
+  PackageDefinition.getLoopBackVersion = function(cb) {
```
bajtos: AFAICT, this is not used anymore, please remove.

common/models/workspace.js
```diff
@@ -183,6 +183,12 @@ module.exports = function(Workspace) {
       if (template.package) {
         template.package.name = packageName;
         template.package.description = description;
+        
+        var loopback = 'loopback';
+        if (loopback in options) {
+          template.package.dependencies[loopback] = options.loopback;
+        }
```
bajtos: In order to switch between loopback 2.x and 3.x, it's not enough to change `dependencies.loopback`. 2.x applications must have `loopback-datasource-juggler` as a top-level dependency too. OTOH, LoopBack 3.0 will finally have juggler as a regular dependency of loopback itself.

```
// loopback 2.x app
"dependencies": {
    "loopback": "^2.0.0",
    "loopback-datasource-juggler": "^2.0.0"
}

// loopback 3.x app
"dependencies": {
    "loopback": "^3.0.0"
}
```

I am proposing to revert this change here and change dependencies only inside tests.

test/model-method.js
```diff
@@ -101,3 +101,112 @@ describe('ModelMethod', function() {
       });
   });
 });
+
+describe('ModelMethod - Loopback 3.0', function() {
+  var userModel;
+
+  beforeEach(function(done) {
+    givenLBXWorkspace({'loopback':'^3.0.0'}, done);
```
bajtos: How about this:

```js
givenWorkspaceWithCustomDependencies = function(template, deps, cb) {
  // ...
}

givenLB2Workspace = function(cb) {
  givenWorkspaceWithCustomDependencies('empty-server', { 
    loopback: '^2.0.0',
    loopback-datasource-juggler: '^2.0.0'
  }, cb);
};

givenLB3Workspace = function(cb) {
  givenWorkspaceWithCustomDependencies('empty-server', {
    loopback: '^3.0.0-alpha.1' 
   }, cb);
};
```

test/model-method.js
```diff
@@ -101,3 +101,112 @@ describe('ModelMethod', function() {
...
+          expect(list[0]).to.not.have.property('isStatic');
+
+          var cfg = new ConfigFile({ path: 'server/models/user.json' });
+          cfg.load(function(err) {
+            if (err) return done(err);
+            var methods = cfg.data.methods;
+            expect(methods).to.be.an('object');
+            expect(methods).to.have.property('testMethod');
+            expect(methods).to.not.have.property('prototype.testMethod');
+            expect(methods).to.not.have.property('isStatic');
```
bajtos: You are checking that "methods" object don't contain `isStatic` flag, I think it's more important to test that method definition does not contain this flag. Thoughts?

```js
expect(methods.testMethod).to.not.have.property('isStatic');
```

common/models/model-definition.json
0candy: `key` has to be set as well otherwise the property doesn't set the name property to the method name.

common/models/model-method.js
0candy: Agreed!

common/models/model-method.js
```diff
@@ -10,4 +11,38 @@ module.exports = function(ModelMethod) {
...
    */
+
+  ModelMethod._shouldEncodeStaticFlagInName = function() {
+    var version = ModelMethod.app.models.Workspace.loopBackVersion;
+    return encodeStaticFlagInName(version);
+  };
+
+  ModelMethod.getJsonKey = function(data) {
+    if (this._shouldEncodeStaticFlagInName()) {
+      var isStatic = data.isStatic;
```
bajtos: In order to reduce nesting/indentation level, it's better to use reverse logic + early return. It also makes the code easier to read, as one can quickly determine what the other branch does and forget about it. In your version, one has to keep in mind that they need to check what happens when static flag is not encoded after they finish examination of the branch when the flag is encoded.

```js
{
  if (!this._shouldEncodeStaticFlagInName())
    return data.name;

  var isStatic = data.isStatic;
  // ...
}
```

common/models/model-method.js
```diff
@@ -10,4 +11,38 @@ module.exports = function(ModelMethod) {
    * @class ModelMethod
    * @inherits WorkspaceEntity
    */
+
+  ModelMethod._shouldEncodeStaticFlagInName = function() {
+    var version = ModelMethod.app.models.Workspace.loopBackVersion;
+    return encodeStaticFlagInName(version);
```
bajtos: nitpick: I would personally inline `encodeStaticFlagInName` here and remove it.

Feel free to ignore this comment though.

test/model-method.js
```diff
@@ -101,3 +101,220 @@ describe('ModelMethod', function() {
...
+        });
+      });
+  });
+  
+  it('add prototype method without isStatic flag to model definition', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'prototype.testMethod',
+        isStatic: false
```
bajtos: The test name says "without isStatic flag", but the test data does set the flag. PTAL

test/model-method.js
```diff
@@ -101,3 +101,220 @@ describe('ModelMethod', function() {
...
+            expect(methods).to.not.have.property('testMethod');
+            expect(methods['prototype.testMethod']).to.not.have.property('isStatic');
+            done();
+          });
+        });
+      });
+  });
+});
+
+function givenWorkspaceWithCustomDependencies(templateName, deps, cb) {
```
bajtos: I think this is generally useful, please move it to `test/support.js`. Consider moving the other two `givenLB*` helpers too.

bajtos:
Almost there :)

test/model-method.js
0candy: I meant isStatic flag should not be included in the method definition. I will reword it.

common/models/model-method.js
```diff
@@ -10,4 +11,35 @@ module.exports = function(ModelMethod) {
...
+    }
+    var isStatic = data.isStatic;
+    if (isStatic !== undefined) {
+      var name = data.name;
+      var matchName = name.match(/^prototype\.(.*)$/);
+      if (!isStatic && (matchName === null || !matchName)) {
+        data.name = 'prototype.' + name;
+      }
+    }
+    delete data.isStatic;
```
bajtos: It's better to have functions as pure as possible, i.e. without side-effects. Especially when the name (`getJsonKey`) does not convey any hint that the method is changing the data passed in the input argument. Please move this line to `getConfigFromData`, even though it means that you have to check for `_shouldEncodeStaticFlagInName` again.

bajtos: Oh, `getConfigFromData` already contains code to remove `config.isStatic`. I think it should be enough to delete this line?

test/model-method.js
```diff
@@ -101,3 +101,186 @@ describe('ModelMethod', function() {
...
+          });
+        });
+      });
+  });
+  
+  it('prototype method should not have isStatic property in method definition', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'prototype.testMethod',
```
bajtos: Hmm, I think this is not valid usage of the API. The name should be always without `prototype.` prefix, at least that's what I thought?

bajtos:
@0candy I just realised that this patch so far deals only with the part where we take `ModelMethod` entity and save it to model JSON file. However, I think we are missing the second part that will handle loading of `ModelMethod` from the JSON file.

I.e. when inside LB 3.0 project, using the example from release notes:

```js
methods: {
  staticMethod: {
    http: { path: '/static' }
  },
  'prototype.instanceMethod': {
  http: { path: '/instance' }
}
```

then the following `ModelMethod` entities should be created:

1. `{ name: 'staticMethod', isStatic: true, ... }`
2. `{name: 'instanceMethod', isStatic: false, ...}`

common/models/model-method.js
```diff
@@ -10,4 +11,58 @@ module.exports = function(ModelMethod) {
...
+    var properties = this.definition.properties;
+    var result = {};
+    var prop;
+
+    // add pre-defined properties in the order defined by LDL
+    // apply `json` config from LDL along the way
+    for (prop in properties) {
+      if (properties[prop].json === false) continue;
+      result[properties[prop].json || prop] = data[prop];
+    }
```
bajtos: Can we please not copy this code and call reuse `ModelMethod.base.getConfigFromData` instead?

common/models/model-method.js
```diff
@@ -10,4 +11,58 @@ module.exports = function(ModelMethod) {
...
+  };
+  
+  ModelMethod.getDataFromConfig = function(config) {
+    var properties = this.definition.properties;
+    
+    if (this._shouldEncodeStaticFlagInName()) {
+      var name = properties.name;
+      var m = name.match(/^prototype\.(.*)$/);
+      var isStatic = !m; 
+      config[name] = isStatic ? suspect : m[1];
```
bajtos: I think `suspect` is not defined, could you please check?

common/models/model-method.js
```diff
@@ -10,4 +11,58 @@ module.exports = function(ModelMethod) {
...
+      config[isStatic] = isStatic;
```
bajtos: This should be `config.isStatic`.

bajtos: Otherwise you are setting `config.true = true` or `config.false = false`.

bajtos:
@0candy 
> a03c484 Code to handle loading of JSON file.

Please add some tests for this too.

common/models/model-method.js
bajtos: > I had to revert this code review change we discussed earlier about using the parent method as it was using the wrong model. When it calls the base method, it uses workspace properties which were different from the modelmethod properties that I wanted.

Oh, I think we need to call the base method like this:

```js
ModelMethod.base.getConfigFromData.call(this, data);
```

common/models/model-definition.json
0candy: `key` needs to be there so it gets removed before writing to JSON.
https://github.com/strongloop/loopback-workspace/pull/273/files#diff-36dfec9040d3e62a2c6ca85e70ab0e7bL176

bajtos: I am confused. You can delete a property that does not exist:

```js
var data = {};
delete data.key;
// works fine, no error is thrown
```

I don't see why we need to keep the `key` around when there is `keyFn`, could you please check again?

common/models/definition.js
```diff
@@ -128,12 +128,13 @@ module.exports = function(Definition) {
               fileData.name, embedId, config);
           }
 
-          config = Entity.getDataFromConfig(config);
-
           // add extra properties for relations
           config[relation.foreignKey] = fk;
           config[relation.embed.key] = embedId;
```
bajtos: Oh, is this the line where we need the `key` property? I find this rather difficult to spot and reason about. Can you please check if there is a solution that is easier to understand? If we cannot find a better way then we can use what you have here, but I'd like you to at least give this a try.

bajtos: How about this?

```js
// ...
config = Entity.getDataFromConfig(config, embedId);

// add extra properties for relations
config[relation.foreignKey] = fk;
if (relation.embed.keySetter) {
  Entity[relation.embed.keySetter](embedId, config)
} else {
  config[relation.embed.key] = embedId;
}
config.facetName = facetName;
// ...
```

Then I would also rename `keyFn` to `keyGetter`.

test/model-method.js
```diff
@@ -249,6 +258,39 @@ describe('ModelMethod - Loopback 3.0', function() {
...
+          });
+        });
+      });
+  });
+
+  it('loading JSON should have correct method name and isStatic flag', function(done) {
+    ModelMethod.create(
+      {
+        modelId: userModel.id,
+        name: 'prototype.testMethod',
```
bajtos: Sorry for repeating myself, but isn't this invalid API usage? IMO, the `name` should be always the short name without `prototype` prefix.

test/model-method.js
```diff
@@ -249,6 +258,39 @@ describe('ModelMethod - Loopback 3.0', function() {
...
+        userModel.methods(function(err, list) {
+          if (err) return done(err);
+          expect(list).to.have.length(1);
+          expect(list[0]).to.have.property('name', 'testMethod');
+          expect(list[0]).to.have.property('isStatic');
+
+          ModelMethod.find(function(err, methods) {
+            if (err) return done(err);
+            expect(methods[0]).to.have.property('name', 'testMethod');
+            expect(methods[0]).to.not.have.property('name', 'prototype.testMethod');
```
bajtos: Again, name should not include "prototype." prefix.

0candy: Yes, this is checking that name `prototype.testMethod` does not exist.

bajtos: Ah, sorry, you are right. This line is correct.

common/models/model-definition.js
```diff
@@ -75,18 +75,18 @@ function ready(ModelDefinition) {
...
         relatedData.forEach(function(related) {
           var Definition = app.models[relation.model];
           var key;
           if (relation.embed.key) {
             key = related[relation.embed.key];
           }
-          var keyFn = relation.embed.keyFn;
-          if (keyFn && typeof Definition[keyFn] === 'function') {
-            key = Definition[keyFn](related);
+          var keySetter = relation.embed.keySetter;
```
bajtos: I think this should be `keyGetter`, because we want to build a key from the data stored in our domain entity, shouldn't it?

common/models/model-definition.js
```diff
@@ -182,8 +182,8 @@ function ready(ModelDefinition) {
     for (var ix in relatedData) {
       var data = Entity.getConfigFromData(relatedData[ix]);
       delete data[relation.foreignKey];
-      delete data[relation.embed.key];
-      delete data[relation.embed.keyFn];
+      delete data.name;
+      delete data[relation.embed.keySetter];
```
bajtos: I don't think we should always delete `data.name`, we should `delete data[relation.embed.key]` instead when the key is set. When it's not set and there is `keySetter`, then it's the responsibility of `getConfigFromData` to remove the `name` property if it is not expected in the output.

bajtos: The idea is to keep `cleanRelatedData` totally not opinionated, just providing extension hooks allowing domain entities to implement whatever custom rules they need.

common/models/model-definition.js
```diff
@@ -75,10 +75,19 @@ function ready(ModelDefinition) {
...
+          'embed as object requires "key" or "keySetter"');
         result = {};
         relatedData.forEach(function(related) {
-          var key = related[relation.embed.key];
+          var Definition = app.models[relation.model];
+          var key;
+          if (relation.embed.key) {
+            key = related[relation.embed.key];
+          }
+          var keySetter = relation.embed.keySetter;
```
bajtos: I think this should be `keyGetter`, because we want to build a key from the data stored in our domain entity, shouldn't it?

The idea in my [comment](https://github.com/strongloop/loopback-workspace/pull/273#discussion_r62665458) was to allow two functions: `keyGetter` to get the key from domain data (cache entry), `keySetter` to set the key in the domain data (cache entry).

Now if we can implement this in such way that only a single function is needed, then that's great! Just use a different name for the single function, I think the original `keyFn` is good enough.

However, I am little bit concerned that the approach of a single method may not be future proof - what if the cached data is very different from the JSON data and the algorithms for get and set are thus different too?

@0candy thoughts?

common/models/model-definition.js
```diff
@@ -173,7 +182,8 @@ function ready(ModelDefinition) {
     for (var ix in relatedData) {
       var data = Entity.getConfigFromData(relatedData[ix]);
       delete data[relation.foreignKey];
-      delete data[relation.embed.key];
+      delete data.name;
```
bajtos: I don't think we should always delete `data.name`, we should `delete data[relation.embed.key]` instead when the key is set. When it's not set and there is `keySetter`, then it's the responsibility of `getConfigFromData` to remove the `name` property if it is not expected in the output.

The idea is to keep `cleanRelatedData` totally not opinionated, just providing extension hooks allowing domain entities to implement whatever custom rules they need.

common/models/model-definition.js
bajtos: @0candy I clarified the first sentence in my comment above:

I think this should be `keyGetter`, because we want to build a key from the data stored in our domain entity (i.e. "get it"), as opposed to setting applying the key to the domain entity (i.e. "set it").

0candy: Yeah, I think we are in agreement, just the naming convention is different. We should have a `keyGetter` and a `keySetter`.
`key` will become `keySetter` and `keyFn` will be `keyGetter`

common/models/definition.js
```diff
@@ -128,12 +128,13 @@ module.exports = function(Definition) {
               fileData.name, embedId, config);
           }
 
-          config = Entity.getDataFromConfig(config);
+          config = Entity.getDataFromConfig(config, embedId);
 
           // add extra properties for relations
           config[relation.foreignKey] = fk;
           config[relation.embed.key] = embedId;
```
bajtos: I think you need to handle the case when `relation.embed.key` is not defined, because there are `embed.keySetter` and `embed.keyGetter` instead.

common/models/model-method.js
```diff
@@ -10,4 +11,48 @@ module.exports = function(ModelMethod) {
...
+    }
+
+    return config;
+  };
+
+  ModelMethod.getDataFromConfig = function(config, name) {
+    var data = ModelMethod.base.getDataFromConfig.call(this, config);
+    data.name = name;
+
+    if (this._shouldEncodeStaticFlagInName()) {
```
bajtos: I see. I was expecting that `keySetter` would be a function that converts `prototype.updateAttributes` key into `name:updateAttributes` & `isStatic:false`, while you are taking care of that conversion directly here, in `getDataFromConfig`. Neat!

common/models/model-definition.json
```diff
@@ -96,7 +96,8 @@
     "methods": {
       "embed": {
         "as": "object",
-        "key": "name"
+        "keySetter": "name",
```
bajtos: Since the conversion from the key in JSON file into name/isStatic in data is already handled by `getDataFromConfig`, `keySetter` is no longer needed and can be removed. Unless I am missing something?

common/models/model-definition.js
```diff
@@ -174,6 +183,8 @@ function ready(ModelDefinition) {
       var data = Entity.getConfigFromData(relatedData[ix]);
       delete data[relation.foreignKey];
       delete data[relation.embed.key];
+      delete data[relation.embed.keySetter];
+      delete data[relation.embed.keyGetter];
```
bajtos: `keySetter` and `keyGetter` are names of functions defined on entity constructor, they shouldn't be stored in instance data. I think these two lines are redundant and should be removed.

I think once you remove these two lines, the `name` property won't be deleted from the config data. In my opinion, the `name` property should be deleted by `ModelMethod.getConfigFromData`. That way the entity (ModelMethod for example) has more control over what's being stored in the config and it should make the code easier to understand to readers that are not entirely familiar with `formatRelatedData`/

bajtos:
@0candy LGTM. Please check the CI build failure on amazon-0.12 - [link](ci.strongloop.com/job/loopback-workspace/label=amazon-0.12/4017/) and squash the commits before landing.

I am going to restart CI to see if the problem was temporary.

bajtos:
@slnode test please

0candy:
amazon-0.12 is not failing. The only thing failing now is generator-loopback and that's waiting to be fixed.

PR #275, opened 2016-04-16
Added "schema" property for the DB2 Connector. Fixes #274.

### LINKED ISSUE #274: DB2 Connector UI/CLI guide does not include "schema" property.
@qpresley I'm not entirely sure if this is necessary, as I haven't used DB2 a lot. However, in my scenario all of my models fall under a particular schema. For instance:

`GOSALES.PRODUCT`

Can I get around setting a schema by naming my model appropriately? Or should we include `schema` [here](https://github.com/strongloop/loopback-workspace/blob/master/available-connectors.json#L22-L57).

Freyert:
The "schema" property is not included for the DB2 connector when creating a connector from the CLI or the UI.

@raymondfeng recommended I include the property here. I'm not sure if there are more properties that should be set, such as it being an optional property (I believe).

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -47,6 +47,9 @@
       },
       "database": {
         "type": "string"
+      },
+      "schema": {
+        "type": "string"
       }
     },
     "package": {


slnode:
Can one of the admins verify this patch? To accept patch and trigger a build add comment ".*ok\W+to\W+test.*"

slnode:
Can one of the admins verify this patch? To accept patch and trigger a build add comment ".*ok\W+to\W+test.*"

Freyert:
Not exactly sure where to place the PR.

bajtos:
Landed, thank you for the contribution!

PR #276, opened 2016-04-18
add strong-error-handler template

loay:


bajtos:
Hey @loay, your pull requests configures loopback-workspace REST API to use the new error handler, but does not modify the actual templates. Because loopback-workspace is never really run in a production mode, the changes you made so far can be reverted.

You need to modify these files instead:
 - https://github.com/strongloop/loopback-workspace/blob/bcb1dc85e3e3d2fcc986ff94e2f4684861c6f3f6/templates/projects/empty-server/data.js
 - https://github.com/strongloop/loopback-workspace/blob/bcb1dc85e3e3d2fcc986ff94e2f4684861c6f3f6/templates/projects/empty-server/files/server/middleware.json
 - https://github.com/strongloop/loopback-workspace/blob/bcb1dc85e3e3d2fcc986ff94e2f4684861c6f3f6/templates/projects/empty-server/files/server/middleware.production.json

server/config.json
```diff
@@ -19,8 +19,8 @@
       "limit": "100kb"
     },
     "cors": false,
-    "errorHandler": {
-      "disableStackTrace": false
+    "strongErrorHandler": {
+      "debug": false
```
bajtos: The keys are referring to middleware module/file names. I think you should use `strong-error-handler` here. Also since loopback-workspace is never used in production-like mode, I think `debug` can be always set to `true`.

bajtos: Ignore my comment above, I thought this was in middleware config. I'll delete both comments.

server/server.js
```diff
@@ -103,7 +103,7 @@ app.use(loopback.urlNotFound());
  */
 
 // The ultimate error handler.
-app.use(loopback.errorHandler());
+app.use(loopback.strongErrorHandler());
```
bajtos: This is weird, apparently the person upgrading this project added JSON files but forgot to change `server/server.js` to match the new template. Let's revert all changes outside `templates` to keep this PR focused.

templates/projects/empty-server/data.js
```diff
@@ -79,8 +79,8 @@ template.server = {
           limit: '100kb'
         },
         cors: false,
-        errorHandler: {
-          disableStackTrace: false
+        strongErrorHandler: {
```
bajtos: I find it difficult to review changes in templates without seeing the corresponding changes in runtime (strong-remoting and/or loopback). Could you please point me to a pull request adding `strongErrorHandler` config option to strong-remoting?

templates/projects/empty-server/files/server/middleware.development.json
```diff
@@ -0,0 +1,9 @@
+{
+  "final:after": {
+    "loopback#strongErrorHandler": {
```
bajtos: Please don't add new middleware to `loopback` namespace, that's an antipattern we have preserved only to simplify migration from LB 1.x (express 3.x) to LB 2.x (express 4.x). In general, people should install `strong-error-handler` directly.

templates/projects/empty-server/files/server/middleware.json
```diff
@@ -47,6 +47,6 @@
     "loopback#urlNotFound": {}
   },
   "final:after": {
-    "loopback#errorHandler": {}
+    "loopback#strongErrorHandler": {}
```
bajtos: ditto

bajtos:
Looks better. As I said, we should load the new module directly and don't depend on loopback deps. In which case you need to modify https://github.com/strongloop/loopback-workspace/blob/bcb1dc85e3e3d2fcc986ff94e2f4684861c6f3f6/templates/projects/empty-server/data.js to include the new dependency.

Issue #272, opened 2016-04-07
Template Naming and Simple Echo Hello World


ritch:
We should rename the current `hello-world` template to "Notes" and create a new simpler example the demonstrates a very basic hello world.

The new hello-world example should only include

 - a single vanilla Message (eventually a controller + model)
 - a single remote method

jannyHou:
- [ ] **Add new template** 
I think adding a new template is similar to https://github.com/strongloop/loopback-workspace/pull/255

- [ ] **Rename old `hello-world` to `Notes`**

@ritch @bajtos could you explain more about the meaning of `controller` in the following requirement? Thanks:

> a single vanilla Message (eventually a controller + model)

bajtos:
> could you explain more about the meaning of controller in the following requirement? Thanks:

> >    a single vanilla Message (eventually a controller + model)

I think the idea is to use a `Message` model for the initial version (now). When we implement support for controllers, the template should be eventually reworked to use controllers.

@ritch can you confirm please?

/cc @richardpringle

richardpringle:
I have an example controller up which I can adjust according to whatever we decide on. It simply echoes a message right now. I will drop a link in here when I finish with some comments.

ritch:
> @ritch can you confirm please?

exactly

jannyHou:
Pr https://github.com/strongloop/loopback-workspace/pull/278 landed.
Will add controller after feature https://github.com/strongloop/loopback-sandbox/pull/7 landed.

ritch:
We should publish loopback-workspace now... and let @kraman know that a new version is out.

@bajtos ^^^

jannyHou:
Verify Steps:
[1] notes template:

- CRUD operations of model `Note` works fine
- Memory db is created in `datasource.json`
- Model `Note` is attached to db

[2] hello-world template:

- remote method `greet` returns response message
- no db in `datasource.json`
- Model `Message` is attached to db null

bajtos:
> We should publish loopback-workspace now... and let @kraman know that a new version is out.

> @bajtos can you do the publish (unless there is anything that would prevent us from publishing?).

@ritch @jannyHou Since this is a user-facing piece, we need somebody from QA to verify this story first. I am moving it from `#review` to `#verify` and reassigning to @cgole.

bajtos:
@jannyHou you wrote:

> **Verified**
> ready to publish

Who has verified the story?

ssh24:
`Verified`

[...]

PR #278, opened 2016-05-04
Add new helloworld

jannyHou:
Connect to strongloop/loopback-workspace#272

- [x] **Add new template** 
Now there is only a `message` model

- [x] **Rename old `hello-world` to `Notes`**

### DIFF
--- a/templates/projects/hello-world/files/common/models/message.json
+++ b/templates/projects/hello-world/files/common/models/message.json
@@ -1,14 +1,6 @@
 {
-  "name": "note",
-  "properties": {
-    "title": {
-      "type": "string",
-      "required": true
-    },
-    "content": {
-      "type": "string"
-    }
-  },
+  "name": "Message",
+  "properties": {},
   "methods": {
     "greet": {
       "isStatic": true,
--- a/templates/projects/hello-world/files/common/models/note.js
+++ b/templates/projects/hello-world/files/common/models/note.js
@@ -1,8 +1,0 @@
-module.exports = function(Note) {
-  Note.greet = function(msg, cb) {
-    process.nextTick(function() {
-      msg = msg || 'world';
-      cb(null, 'Hello ' + msg);
-    });
-  };
-};
--- a/templates/projects/notes/files/common/models/note.js
+++ b/templates/projects/notes/files/common/models/note.js
@@ -1,0 +1,2 @@
+module.exports = function(Note) {
+};
--- a/templates/projects/notes/files/common/models/note.json
+++ b/templates/projects/notes/files/common/models/note.json
@@ -1,0 +1,12 @@
+{
+  "name": "Note",
+  "properties": {
+    "title": {
+      "type": "string",
+      "required": true
+    },
+    "content": {
+      "type": "string"
+    }
+  }
+}
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -324,6 +324,82 @@ describe('end-to-end', function() {
     });
   });
 
+  describe('notes template', function() {
+    var app, modelInstance;
+
+    before(resetWorkspace);
+    before(givenEmptySandbox);
+
+    before(function createWorkspace(done) {
+      givenWorkspaceFromTemplate('notes', done);
+    });
+
+    before(installSandboxPackages);
+
+    before(function loadApp() {
+      app = require(SANDBOX);
+    });
+
+    it('provides retrieve operation', function(done) {
+      request(app)
+        .get('/api/Notes')
+        .expect(200, done);
+    });
+
+    it('provides create operation', function(done) {
+      var sample = {title: 'myTitle'};
+      request(app)
+      .post('/api/Notes')
+      .send(sample)
+      .expect(200, function(err, res) {
+        if(err) {
+          done(err)
+        } else {
+          expect(res.body).
+            to.have.property('title', 'myTitle');
+          done();
+        };
+      });
+    });
+
+    it('provides update operation', function(done) {
+      var sample = {title: 'myTitle'};
+      request(app)
+      .put('/api/Notes')
+      .send(sample)
+      .expect(200, function(err, res) {
+        if(err) {
+          done(err)
+        } else {
+          expect(res.body).
+            to.have.property('title', 'myTitle');
+          done();
+        };
+      });
+    });
+
+    it('provides delete operation', function(done) {
+      var Note = app.models.Note;
+      Note.create({title: 'myTitle'}, function(error, note) {
+        if(error) {
+          done(error);
+        } else {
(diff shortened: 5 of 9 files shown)


jannyHou:
@bajtos So in this pr I just add one model `message` to new template `hello-world`.
And rename the original `hello-world` to `notes`, made a little change of the greeting message of model `note`, modified the test case as well.
Please review. Thanks.

jannyHou:
Seems it breaks lots of test cases in generator-loopback. I will take a look into it then.

templates/projects/notes/files/common/models/note.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function(Note) {
+  Note.greet = function(msg, cb) {
+    process.nextTick(function() {
+      msg = msg || 'Note';
+      cb(null, 'Model ' + msg);
+    });
+  };
```
bajtos: @ritch we are cleaning up example projects offered by `apic loopback`, as you requested in https://github.com/strongloop/loopback-workspace/issues/272.

I think the `Notes` project should not include any custom remote methods, as it should demonstrate CRUD/data access API. What do you think?

If you prefer to keep a custom remote method, then I think it should be definitely something else than `Note.greet`. What would be a better method?

test/end-to-end.js
```diff
@@ -324,14 +324,14 @@ describe('end-to-end', function() {
     });
   });
 
-  describe('hello-world template', function() {
+  describe('notes template', function() {
```
bajtos: I'd like to have both `notes` and `hello-world` templates covered by these tests.

bajtos:
Looks mostly good, let's wait for Ritchie to comment on https://github.com/strongloop/loopback-workspace/pull/278#discussion_r62000274

test/end-to-end.js
jannyHou: Yeah I will add test for new template as well.

templates/projects/notes/files/common/models/note.js
ritch: > I think the Notes project should not include any custom remote methods, as it should demonstrate CRUD/data access API. What do you think?

I agree... It should just be CRUD.

jannyHou:
@bajtos I added CRUD test cases for template `note` and verify the endpoint of modle message works in template `hello-world`.
Please review:) Thanks.

Maybe you can go to template folder to see if templates are defined properly, label `File Changes` looks messy....

templates/projects/hello-world/files/common/models/message.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "message",
```
bajtos: Nitpick: we usually use `PascalCase` for model names.

bajtos: Also the template description mentions `Message`, while you are using `message` here.

templates/projects/notes/files/common/models/note.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "note",
```
bajtos: Ditto, it will be better to use `Note` as the name. (I do realise the original template used `note` instead.)

templates/projects/hello-world/files/common/models/message.json
```diff
@@ -0,0 +1,12 @@
...
+  "properties": {
+    "title": {
```
bajtos: I am not sure if it makes sense to specify `title` and `content` properties of the `Message` model, when we are not using them at all?

templates/projects/hello-world/files/common/models/message.json
bajtos: One more thing: since the Message model is not persisted in any datasource, it should specify `"base": "Model"` so that CRUD methods are not included in the REST API.

bajtos:
@jannyHou in [templates/projects/hello-world/data.js](https://github.com/strongloop/loopback-workspace/blob/0e6b3c2a5f916011f5144c4352cacba18a7c1b3a/templates/projects/hello-world/data.js#L26), the model `Message` should not be attached to any datasource, i.e. `"dataSource": null`. The default datasource `db` should be removed from the template too, see https://github.com/strongloop/loopback-workspace/blob/0e6b3c2a5f916011f5144c4352cacba18a7c1b3a/templates/projects/hello-world/data.js#L31-L34

templates/projects/hello-world/files/common/models/note.js
```diff
@@ -1,8 +0,0 @@
-module.exports = function(Note) {
-  Note.greet = function(msg, cb) {
-    process.nextTick(function() {
-      msg = msg || 'world';
-      cb(null, 'Hello ' + msg);
-    });
-  };
```
bajtos: AFAICT, the greet method was removed completely. Instead, it should be added to the `Message` model.

https://github.com/strongloop/loopback-workspace/blob/0e6b3c2a5f916011f5144c4352cacba18a7c1b3a/templates/projects/hello-world/files/common/models/message.js

And the remoting metadata should go here:
https://github.com/strongloop/loopback-workspace/blob/0e6b3c2a5f916011f5144c4352cacba18a7c1b3a/templates/projects/hello-world/files/common/models/message.json

test/end-to-end.js
```diff
@@ -340,15 +416,10 @@ describe('end-to-end', function() {
...
+    it('contains model "message"', function(done) {
       request(app)
-        .get('/api/notes/greet?msg=Tester')
-        .expect(200, function(err, res) {
-          if (err) done(err);
-          expect(res.body)
-            .to.have.property('greeting', 'Hello Tester');
-          done();
-        });
+        .get('/api/messages')
```
bajtos: The functionality described in this test should be preserved, just the endpoint should be different (`GET /api/messages/greet?msg=Tester`).

jannyHou:
@bajtos I applied your feedback:
1. Change model name to PascalCase (`Note` and `Message`)
2. Attach `Message` to no db, delete db in template note, add a single remote method to `Message`
3. > I am not sure if it makes sense to specify title and content properties of the Message model, when we are not using them at all?

`title` and `content` are for model `note`
I defined two properties called `sender` and `receiver` to make it like a message. And in remote method, I defined the greeting message as 
```js
'Sender says ' + msg + ' to receiver'
```
But if you feel it not useful to have properties unused in template, then let's delete `sender` and `receiver` too lol.

bajtos:
>  I defined two properties called sender and receiver to make it like a message.
> But if you feel it not useful to have properties unused in template, then let's delete sender and receiver too lol.

It's better to avoid adding stuff that's not used. Let's remove these two properties please.

Other than that, the patch LGTM. No further review is necessary, please squash the commits before landing.

PR #281, opened 2016-05-26
Fix Message extends Model

jannyHou:
A fix to https://github.com/strongloop/loopback-workspace/pull/278
@bajtos Sorry I realized that I didn't specify `Message` to extend `Model` instead of `PersistedModel` in the new hello-world template.

I fixed it in this pr. Please review, thanks!

### DIFF
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -426,6 +426,12 @@ describe('end-to-end', function() {
           done();
         });
     });
+
+    it('comes without built-in GET endpoint', function(done) {
+      request(app)
+        .get('/api/Messages')
+        .expect(404, done);
+    });
   });
 
   describe('autoupdate', function() {
--- a/templates/projects/hello-world/files/common/models/message.json
+++ b/templates/projects/hello-world/files/common/models/message.json
@@ -1,5 +1,6 @@
 {
   "name": "Message",
+  "base": "Model",
   "properties": {},
   "methods": {
     "greet": {


bajtos:
@jannyHou Please add a unit-test to verify that the `Message` model does not provide CRUD endpoints in the REST API.

jannyHou:
@bajtos I added test to verify that REST apis for `Message` are hidden. please review:) Thanks.

test/end-to-end.js
```diff
@@ -426,6 +426,30 @@ describe('end-to-end', function() {
           done();
         });
     });
+
+    it('hides endpoint GET', function(done) {
```
bajtos: I think "hides" may be a bit misleading, since there is no GET endpoint at all. Perhaps this name would be better?

```js
it('comes without built-in GET endpoint')
```

test/end-to-end.js
```diff
@@ -426,6 +426,30 @@ describe('end-to-end', function() {
...
+      request(app)
+        .get('/api/Messages')
+        .expect(404, done);
+    });
+
+    it('hides endpoint POST', function(done) {
```
bajtos: I am not sure if it's necessary to test four endpoints. Why exactly four tests, why not only two, or why not one test for each PersistedModel route? I would personally keep the first test only.

However, if you feel it's better to have more tests, then I am ok with that too, just apply the comment above to names.

bajtos:
@jannyHou no more reviews are needed, please squash the commits before landing.

PR #283, opened 2016-05-31
Choose dependency version according to loopbackVersion

jannyHou:
Connect to strongloop/loopback#2344

### DIFF
--- a/common/models/workspace.js
+++ b/common/models/workspace.js
@@ -33,6 +33,14 @@ module.exports = function(Workspace) {
     var ViewDefinition = app.models.ViewDefinition;
     var TEMPLATE_DIR = path.join(__dirname, '..', '..', 'templates', 'projects');
     var DEFAULT_TEMPLATE = 'api-server';
+    var DEFAULT_LB_VERSION = '2.x';
+    var DEPENDENCIES_3_X = {
+      'loopback': '^3.0.0-alpha.1',
+    };
+    var DEPENDENCIES_2_X = {
+      'loopback': '^2.22.0',
+      'loopback-datasource-juggler': '^2.39.0',
+    };
     var debug = require('debug')('workspace');
 
     /**
@@ -152,6 +160,7 @@ module.exports = function(Workspace) {
       if (!options.root) {
         throw new Error('Non-root components are not supported yet.');
       }
+      var loopbackVersion = options.loopbackVersion || DEFAULT_LB_VERSION;
       var templateName = options.template || DEFAULT_TEMPLATE;
       var name = options.name || templateName;
       var packageName = options.packageName || name;
@@ -168,6 +177,14 @@ module.exports = function(Workspace) {
         err.statusCode = 400;
         return cb(err);
       }
+
+      if (loopbackVersion !== '2.x' && loopbackVersion !== '3.x') {
+        return cb(new Error('Loopback version should be either 2.x or 3.x'));
+      }
+      var defaultDependencies = template.package.dependencies;
+      var loopbackDependencies =
+        loopbackVersion === '2.x' ? DEPENDENCIES_2_X : DEPENDENCIES_3_X;
+      template.package.dependencies = extend(defaultDependencies, loopbackDependencies);
 
       // TODO(bajtos) come up with a more generic approach
       var explorer = 'loopback-component-explorer';
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -11,6 +11,7 @@ var install = require('strong-cached-install');
 var mysql = require('mysql');
 var path = require('path');
 var request = require('supertest');
+var semver = require('semver');
 var debug = require('debug')('test:end-to-end');
 var workspace = require('../');
 var models = workspace.models;
@@ -347,13 +348,13 @@ describe('end-to-end', function() {
     });
 
     it('provides create operation', function(done) {
-      var sample = {title: 'myTitle'};
+      var sample = { title: 'myTitle' };
       request(app)
       .post('/api/Notes')
       .send(sample)
       .expect(200, function(err, res) {
-        if(err) {
-          done(err)
+        if (err) {
+          done(err);
         } else {
           expect(res.body).
             to.have.property('title', 'myTitle');
@@ -363,13 +364,13 @@ describe('end-to-end', function() {
     });
 
     it('provides update operation', function(done) {
-      var sample = {title: 'myTitle'};
+      var sample = { title: 'myTitle' };
       request(app)
       .put('/api/Notes')
       .send(sample)
       .expect(200, function(err, res) {
-        if(err) {
-          done(err)
+        if (err) {
+          done(err);
         } else {
           expect(res.body).
             to.have.property('title', 'myTitle');
@@ -380,15 +381,15 @@ describe('end-to-end', function() {
 
     it('provides delete operation', function(done) {
       var Note = app.models.Note;
-      Note.create({title: 'myTitle'}, function(error, note) {
-        if(error) {
+      Note.create({ title: 'myTitle' }, function(error, note) {
+        if (error) {
           done(error);
         } else {
           request(app)
           .delete('/api/Notes/' + note.id)
           .expect(200, function(err, res) {
-            if(err) {
-              done(err)
+            if (err) {
+              done(err);
             } else {
               expect(res.body).
                 to.have.property('count', 1);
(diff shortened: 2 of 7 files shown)


jannyHou:
@slnode test please

jannyHou:
@bajtos I assume the difference between 2.x and 3.x are versions of `loopback` and `loopback-datasource-juggler`. So I detect which version to use in `workspace.js`, and the default one is 2.x:

```
    var DEPENDENCY_3_X = {
       'loopback': '^3.0.0-alpha.1',
       'loopback-datasource-juggler': '^3.0.0-alpha.1',
     };
     var DEPENDENCY_2_X = {
       'loopback': '^2.22.0',
       'loopback-datasource-juggler': '^2.39.0',
     };
```

There is a few linter fix in this pr.
The implementation is in [`common/models/workspace.js`](https://github.com/strongloop/loopback-workspace/pull/283/files#diff-7bbb877b22354ec23e89bcd8cf629ed5R36)
And test case:
https://github.com/strongloop/loopback-workspace/pull/283/files#diff-a2a7176a039a956670f80ee8b168d36eR437

common/models/workspace.js
```diff
@@ -33,6 +33,15 @@ module.exports = function(Workspace) {
     var ViewDefinition = app.models.ViewDefinition;
     var TEMPLATE_DIR = path.join(__dirname, '..', '..', 'templates', 'projects');
     var DEFAULT_TEMPLATE = 'api-server';
+    var DEFAULT_LB_VERSION = '2.x';
+    var DEPENDENCY_3_X = {
+      'loopback': '^3.0.0-alpha.1',
+      'loopback-datasource-juggler': '^3.0.0-alpha.1',
```
bajtos: In 3.0, juggler is a regular dependency of loopback, it should not be saves in application dependencies.

bajtos:
Please update the test helpers (`givenLB2Workspace` and `givenLB3Workspace`) to use this new feature too, see https://github.com/strongloop/loopback-workspace/blob/master/test/support.js#L127-L138. I think we can get rid of `givenWorkspaceWithCustomDependencies` afterwards.

common/models/workspace.js
jannyHou: Ok I see, so could you provide more details about what is the entire `dependencies` in LB 3.x's `package.json`
Is it 
```json
'dependencies': {
    'compression': '^1.0.3',
    'cors': '^2.5.2',
    'helmet': '^1.3.0',
    'loopback': '^3.0.0-alpha.1',
    'loopback-boot': '^2.6.5',
    'loopback-component-explorer': '^2.4.0',
    'serve-favicon': '^2.0.1',
  },
```
?

jannyHou:
@bajtos I applied your feedback. Please review:) Thanks.

common/models/workspace.js
bajtos: Uff, I don't know what's the full list of dependencies. Here are the differences between 2.x and 3.x:

```diff
- "loopback": "^2.22.0",
- "loopback-datasource-juggler": "^2.39.0",
+ "loopback": "^3.0.0-alpha.1",
```

See how `givenLB3Workspace` and `givenLB2Workspace` helpers work, they should be already configuring the right set of dependencies.

common/models/workspace.js
```diff
@@ -169,6 +178,17 @@ module.exports = function(Workspace) {
         return cb(err);
       }
 
+      var defaultDependencies = template.package.dependencies;
+      if (loopbackVersion === '2.x') {
+        template.package.dependencies =
+          extend(defaultDependencies, DEPENDENCY_2_X);
+      } else if (loopbackVersion === '3.x') {
+        template.package.dependencies =
+          extend(defaultDependencies, DEPENDENCY_3_X);
```
bajtos: Can we reduce duplication in these two branches? E.g.

```js
var loopbackDependencies = loopbackVersion === '2.x' ? DEPENDENCY_2_X : DEPENDENCY_3_X;
template.package.dependencies = extend(defaultDependencies, loopbackDependencies);
```

Also as I was writing this, it occurred to me that these two constants should use plural in their names, because they can contain more than one dependency. I mean `DEPENDENCIES_2_X` and `DEPENDENCIES_3_X`.

common/models/workspace.js
```diff
@@ -169,6 +178,17 @@ module.exports = function(Workspace) {
...
+      } else {
+        throw new Error('Loopback version should be either 2.x or 3.x');
```
bajtos: Should we return the error via callback instead?

test/end-to-end.js
```diff
@@ -434,6 +434,59 @@ describe('end-to-end', function() {
...
+    before(resetWorkspace);
+    before(function createWorkspace(done) {
+      var options = { loopbackVersion: '3.x' };
+      givenWorkspaceFromTemplate('empty-server', options, done);
+    });
+
+    it('contains dependencies with 3.x version', function(done) {
+      var dependencies = readPackageJsonSync().dependencies;
+      expect(dependencies.loopback).to.equal('^3.0.0-alpha.1');
+      // expect(dependencies['loopback-datasource-juggler'])
```
bajtos: Please remove commented-out code.

test/model-method.js
```diff
@@ -105,7 +105,10 @@ describe('ModelMethod', function() {
 describe('ModelMethod - Loopback 2.0', function() {
   var userModel;
 
-  beforeEach(givenLB2Workspace);
```
bajtos: Could you please preserve `givenLB2Workspace` and `givenLB3Workspace` helpers? IMO it makes the code easier to read by hiding implementation details that are not important.

jannyHou:
@bajtos thank you I applied your feedback, please review it.

common/models/workspace.js
```diff
@@ -178,16 +178,13 @@ module.exports = function(Workspace) {
...
-      if (loopbackVersion === '2.x') {
-        template.package.dependencies =
-          extend(defaultDependencies, DEPENDENCY_2_X);
-      } else if (loopbackVersion === '3.x') {
-        template.package.dependencies =
-          extend(defaultDependencies, DEPENDENCY_3_X);
-      } else {
-        throw new Error('Loopback version should be either 2.x or 3.x');
+      if (loopbackVersion !== '2.x' && loopbackVersion !== '3.x') {
+        cb(new Error('Loopback version should be either 2.x or 3.x'));
```
bajtos: Missing `return`.

```js
return cb(new Error(...));
```

Please add a unit-test covering this branch.

test/end-to-end.js
```diff
@@ -434,6 +434,57 @@ describe('end-to-end', function() {
...
+  describe('scaffold 3.x loopback project with option 3.x', function(done) {
+    before(resetWorkspace);
+    before(function createWorkspace(done) {
+      var options = { loopbackVersion: '3.x' };
+      givenWorkspaceFromTemplate('empty-server', options, done);
+    });
+
+    it('contains dependencies with 3.x version', function(done) {
+      var dependencies = readPackageJsonSync().dependencies;
+      expect(dependencies.loopback).to.equal('^3.0.0-alpha.1');
```
bajtos: This test is brittle, it will start failing when we change the scaffolded loopback version to `3.0.0`. Please rework it to use [semver](https://www.npmjs.com/package/semver)'s range matcher instead. Check how @0candy detects whether workspace contains 2.x or 3.x codebase and use a similar approach.

test/end-to-end.js
```diff
@@ -434,6 +434,57 @@ describe('end-to-end', function() {
...
+    before(function createWorkspace(done) {
+      var options = { loopbackVersion: '2.x' };
+      givenWorkspaceFromTemplate('empty-server', options, done);
+    });
+
+    it('contains dependencies with 2.x version', function(done) {
+      var dependencies = readPackageJsonSync().dependencies;
+      expect(dependencies.loopback).to.equal('^2.22.0');
+      expect(dependencies['loopback-datasource-juggler'])
+        .to.equal('^2.39.0');
```
bajtos: Ditto.

test/end-to-end.js
```diff
@@ -434,6 +434,57 @@ describe('end-to-end', function() {
...
+    before(resetWorkspace);
+    before(function createWorkspace(done) {
+      givenWorkspaceFromTemplate('empty-server', done);
+    });
+
+    it('contains dependencies with 2.x version', function(done) {
+      var dependencies = readPackageJsonSync().dependencies;
+      expect(dependencies.loopback).to.equal('^2.22.0');
+      expect(dependencies['loopback-datasource-juggler'])
+        .to.equal('^2.39.0');
```
bajtos: ditto

test/support.js
```diff
@@ -107,34 +107,19 @@ givenWorkspaceFromTemplate = function(template, options, cb) {
...
-      ConfigFile.loadFromPath('/package.json', function(err, pkgFile) {
-        if (err) return cb(err);
-        for (dep in deps) {
-          pkgFile.data.dependencies[dep] = deps[dep];
-        }
-
-        pkgFile.save(cb());
-      });
-    });
+    givenWorkspaceFromTemplate('empty-server', cb);
```
bajtos: Please be explicit and set `options.loopbackVersion`. We will change the default from 2.x to 3.x in the near future, this helper should continue working even after that change.

bajtos:
@jannyHou reviewed, PTAL ^^^

jannyHou:
@bajtos Feedback applied, please review:) thanks.

bajtos:
LGTM.

PR #284, opened 2016-06-06
Add strongErrorHandler templates

loay:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/896


loay:
@bajtos  PTAL

server/config.json
```diff
@@ -19,8 +19,8 @@
       "limit": "100kb"
     },
     "cors": false,
-    "errorHandler": {
-      "disableStackTrace": false
+    "strong-error-handler": {
+      "isDebugMode": false
```
bajtos: I believe the `errorhandler` section of `remoting` config was not changed, pelase revert.

@davidcheung could you please confirm?

templates/projects/empty-server/data.js
```diff
@@ -80,8 +80,8 @@ template.server = {
         limit: '100kb',
       },
       cors: false,
-      errorHandler: {
-        disableStackTrace: false,
+      strong-error-handler: {
+        isDebugMode: false,
```
bajtos: ditto

templates/projects/empty-server/files/server/middleware.development.json
```diff
@@ -0,0 +1,9 @@
+{
+  "final:after": {
+    "strong-error-handler": {
+      "params": {
+        "isDebugMode": true
```
bajtos: I believe the option is called `debug`, not `isDebugMode`.

@davidcheung could you please confirm?

@loay please add unit-tests to verify the behaviour. It's best to avoid modification of `process.env`, I think this may work well for you:

```js
var app = loopback({ localRegistry: true, loadBuiltinModels: true });
var bootOptions = {
  appRootDir: // path to scaffolded app
  env: 'development' // or 'production'
};
boot(app, bootOptions, function(err) {
  // TODO: handle error
  app.listen(0, function() {
    // make bad request to app.get('url') 
    // and check whether response contains stack trace
  });
});
```

bajtos:
@loay also check why CI is failing.

server/config.json
davidcheung: this is for strong-remoting, I think keeping it as `errorHandler` might be easiest for users 
https://github.com/strongloop/strong-remoting/blob/b03dfe8ec53d4ca694a2e438d6b5dadff0824305/lib/rest-adapter.js#L316
```js
"errorHandler" : {
  "debug": true,
  "log": false
}
```

loay:
@bajtos I fixed the above mentioned issue. I will add the test case now. 
As for the ci failures, the same test cases are also failing for the master branch

PR #285, opened 2016-06-09
update errorHandler templates

loay:
connect to https://github.com/strongloop-internal/scrum-loopback/issues/896

### DIFF
--- a/server/config.json
+++ b/server/config.json
@@ -20,7 +20,8 @@
     },
     "cors": false,
     "errorHandler": {
-      "disableStackTrace": false
+      "debug": true,
+      "log": true
     }
   },
   "legacyExplorer": false
--- a/server/middleware.json
+++ b/server/middleware.json
@@ -17,6 +17,6 @@
     "loopback#urlNotFound": {}
   },
   "final:after": {
-    "errorhandler": {}
+    "strong-error-handler": {}
   }
 }
--- a/templates/projects/empty-server/data.js
+++ b/templates/projects/empty-server/data.js
@@ -78,9 +78,7 @@ template.server = {
         limit: '100kb',
       },
       cors: false,
-      errorHandler: {
-        disableStackTrace: false,
-      },
+      handleErrors: false,
     }},
     { name: 'legacyExplorer', value: false },
   ],
--- a/templates/projects/empty-server/files/server/middleware.json
+++ b/templates/projects/empty-server/files/server/middleware.json
@@ -13,7 +13,7 @@
     },
     "helmet#xssFilter": {},
     "helmet#frameguard": {
-      "params": [ 
+      "params": [
         "deny"
        ]
     },
@@ -26,7 +26,7 @@
     "helmet#hidePoweredBy": {},
     "helmet#ieNoOpen": {},
     "helmet#noSniff": {},
-    "helmet#noCache": { 
+    "helmet#noCache": {
       "enabled": false
     }
   },
@@ -47,6 +47,6 @@
     "loopback#urlNotFound": {}
   },
   "final:after": {
-    "loopback#errorHandler": {}
+    "strong-error-handler": {}
   }
 }
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -19,6 +19,11 @@ var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
 var should = require('chai').should();
+// This may not work when the scaffolded application uses a different
+// loopback or loopback-boot version than the one used in loopback-workspace
+// Please make sure that versions are conform.
+var loopback = require('loopback');
+var boot = require('loopback-boot');
 
 var Workspace = workspace.models.Workspace;
 
@@ -982,6 +987,51 @@ describe('end-to-end', function() {
     }
   });
 });
+
+  it('includes sensitive error details in development mode', function(done) {
+    var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+    var bootOptions = {
+      appRootDir: SANDBOX + '/server',
+      env: 'development'
+    };
+    boot(app, bootOptions, function(err) {
+        request(app)
+          .get('/url-does-not-exist')
+          .expect(404)
+          .end(function(err, res) {
+            if (err) return done (err);
+            var responseBody = JSON.stringify(res.body);
+            expect(responseBody).to.include('stack');
+
+            done();
+        });
+    });
+  });
+
+  it('omits sensitive error details in production mode', function(done) {
+    var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+    var bootOptions = {
+      appRootDir: SANDBOX + '/server',
+      env: 'production'
+    };
+    boot(app, bootOptions, function(err) {
+        request(app)
+          .get('/url-does-not-exist')
+          .expect(404)
+          .end(function(err, res) {
+            // Assert that the response body does not contain stack trace.
+            // We want the assertion to be robust and keep working even
+            // if the property name storing stack trace changes in the future,
+            // therefore we test full response body.
+             if (err) return done(err);
+             var responseBody = JSON.stringify(res.body);
+             expect(responseBody).to.not.have.property('stack');
+             expect(responseBody).to.not.include('stack');
+             expect(JSON.stringify(res.body, null, 2)).to.not.contain(__filename);
+             done();
+        });
+    });
+  });
 
 function setupConnection(done) {
   var connection = mysql.createConnection({
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "async": "^1.5.2",
     "cookie-parser": "^1.3.2",
     "debug": "^2.2.0",
-    "errorhandler": "^1.1.1",
+    "strong-error-handler": "^1.0.1",
     "fs-extra": "^0.30.0",
     "glob": "^7.0.0",
     "lodash": "^4.5.1",
--- a/server/server.js
+++ b/server/server.js
@@ -8,7 +8,11 @@ var path = require('path');
 var methodOverride = require('method-override');
 var app = module.exports = loopback();
 var boot = require('loopback-boot');
+<<<<<<< HEAD
 var cookieParser = require('cookie-parser');
+=======
+var errorHandler = require('strong-error-handler');
+>>>>>>> 7143d31... update errorHandler templates
 
 app.set('legacyExplorer', false);
 
@@ -104,8 +108,7 @@ app.use(loopback.urlNotFound());
  */
 
 // The ultimate error handler.
-app.use(loopback.errorHandler());
-
+app.use(errorHandler());
 /*
  * 5. Add a basic application status route at the root `/`.
  *
--- a/templates/projects/empty-server/files/server/middleware.development.json
+++ b/templates/projects/empty-server/files/server/middleware.development.json
@@ -1,0 +1,10 @@
+{
+  "final:after": {
+    "strong-error-handler": {
+      "params": {
+        "debug": true,
+        "log": true
+      }
+    }
+  }
+}
--- a/templates/projects/empty-server/files/server/middleware.production.json
+++ b/templates/projects/empty-server/files/server/middleware.production.json
@@ -1,9 +1,0 @@
-{
-  "final:after": {
-    "loopback#errorHandler": {
-      "params": {
-        "includeStack": false
-      }
-    }
-  }
-}


loay:
@slnode test please

loay:
@bajtos PTAL
I closed the previous PR: https://github.com/strongloop/loopback-workspace/pull/284 as it seems to be corrupted. Let's use this one instead please. Thanks.

server/config.json
```diff
@@ -20,7 +20,8 @@
     },
     "cors": false,
     "errorHandler": {
-      "disableStackTrace": false
+      "debug": false,
```
bajtos: `disableStackTrace:false` should be converted into `debug:true`.

server/middleware.json
```diff
@@ -17,6 +17,5 @@
     "loopback#urlNotFound": {}
   },
   "final:after": {
-    "errorhandler": {}
```
bajtos: Use `strong-error-handler`.

templates/projects/empty-server/data.js
```diff
@@ -79,7 +79,8 @@ template.server = {
       },
       cors: false,
       errorHandler: {
-        disableStackTrace: false,
+        debug: false,
```
bajtos: `disableStackTrace:false` should be converted into `debug:true`.

templates/projects/empty-server/files/server/middleware.json
```diff
@@ -47,6 +47,5 @@
     "loopback#urlNotFound": {}
   },
   "final:after": {
-    "loopback#errorHandler": {}
```
bajtos: Use strong-error-handler.

bajtos: Also add `server/middleware.development.json`, see https://github.com/strongloop/loopback/pull/2411#discussion_r66607554

test/errorhandler.js
```diff
@@ -0,0 +1,36 @@
...
+var request = require('supertest');
+var util = require('util');
+var async = require('async');
+var app = require('../');
+var loopback = require('loopback');
+var errorHandler = require('strong-error-handler');
+var boot = require('loopback-boot');
+
+describe('loopback.errorHandler(options)', function() {
+  it('should delete stack when options.debug is false', function(done) {
```
bajtos: Honestly, I don't see much value in testing how loopback-workspace treat HTTP error responses.

We need a unit-test verifying behaviour of scaffolded apps, as I described here: https://github.com/strongloop/loopback-workspace/pull/284#discussion_r66055596

bajtos: You need two tests - one with `env: 'development'` and another with `env: 'production'`, they should be added to [test/end-to-end.js](https://github.com/strongloop/loopback-workspace/blob/cde85da8897425c0f4716e46c5bf9c95865df6a3/test/end-to-end.js#L35-L155) and `appRootDir` should be set to `SANDBOX`.

bajtos:
@loay reviewed, PTAL

server/middleware.json
loay: That was used initially, then we agreed to use `errorhandler` instead. .. reverting back now. Fixed.

templates/projects/empty-server/files/server/middleware.json
loay: That was used initially, then we agreed to use `errorhandler` instead. .. reverting back now. Fixed. 
Also, same for `middleware.development.json` .. file was there then was removed since we didn't need the errorhandler middleware anymore. I reverted and fixed the issue. 
All in place now.

loay:
@bajtos PTAL

test/end-to-end.js
```diff
@@ -19,6 +19,8 @@ var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
 var should = require('chai').should();
+var loopback = require('loopback');
+var boot = require('loopback-boot');
```
bajtos: Hmm, I think this may not work when the scaffolded application uses a different loopback or loopback-boot version than we use in loopback-workspace. Considering that fix is not straightforward and the situation may actually never happen, I think it's ok to leave these two lines as they are, but please add a comment pointing out this possible problem.

test/end-to-end.js
```diff
@@ -970,6 +972,36 @@ describe('end-to-end', function() {
...
     });
 
+    it('passes scaffolded tests in development mode', function(done) {
+      var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+      var bootOptions = {
+        appRootDir: SANDBOX,
+        env: 'development'
+      };
+      boot(app, bootOptions, function(err) {
+        app.listen(0, function() {
```
bajtos: Unhandled `err`

test/end-to-end.js
```diff
@@ -970,6 +972,36 @@ describe('end-to-end', function() {
       });
     });
 
+    it('passes scaffolded tests in development mode', function(done) {
```
bajtos: We are not interested in scaffolded tests, but want to see what an error response looks like. To do that, you need to request a URL that returns an error response.

test/end-to-end.js
```diff
@@ -970,6 +972,36 @@ describe('end-to-end', function() {
...
+      var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+      var bootOptions = {
+        appRootDir: SANDBOX,
+        env: 'development'
+      };
+      boot(app, bootOptions, function(err) {
+        app.listen(0, function() {
+          execNpm(['test'], { cwd: SANDBOX }, function(err, stdout, stderr) {
+            done(err);
```
bajtos: Here you need to make that request I mentioned in my earlier comment. See also what I wrote in https://github.com/strongloop/loopback-workspace/pull/284#discussion_r66055596.

test/end-to-end.js
```diff
@@ -970,6 +972,36 @@ describe('end-to-end', function() {
...
+      });
+    });
+
+    it('passes scaffolded tests in production mode', function(done) {
+      var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+      var bootOptions = {
+        appRootDir: SANDBOX,
+        env: 'production'
+      };
+      boot(app, bootOptions, function(err) {
```
bajtos: unhandled `err`

loay:
updated test cases and rebased. 
PTAL @bajtos

test/end-to-end.js
```diff
@@ -19,6 +19,12 @@ var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
 var should = require('chai').should();
+// this may not work when the scaffolded application uses a different
```
bajtos:
```js
// This...
```

to make this consistent with the third line of the comment, which starts with capital letter too

test/end-to-end.js
```diff
@@ -19,6 +19,12 @@ var TestDataBuilder = require('./helpers/test-data-builder');
...
+//loopback or loopback-boot version than the one used in loopback-workspace
+//Please make sure that versions are conform.
```
bajtos: Please add a space between double-slash and the comment text:

```js
// This ...
// loopback...
// Please...
```

test/end-to-end.js
```diff
@@ -970,6 +976,44 @@ describe('end-to-end', function() {
...
+        appRootDir: SANDBOX,
+        env: 'development'
+      };
+      boot(app, bootOptions, function(err) {
+        if (err) return done(err);
+        app.listen(0, function() {
+          request(app)
+            .get('/url-does-not-exist')
+            .end(function(err, res) {
+              expect(res.error.text.match(/url-does-not-exist/));
```
bajtos: Hmm, I find `res.error.text` a bit unexpected, I believe strong-error-handler returns different JSON properties. @davidcheung could you PTAL too?

test/end-to-end.js
```diff
@@ -970,6 +976,44 @@ describe('end-to-end', function() {
       });
     });
 
+    it('shows error behavior responses in development mode', function(done) {
```
bajtos: The test name should describe what is the expected behaviour/outcome. E.g. "it includes sensitive error details in development mode".

test/end-to-end.js
```diff
@@ -970,6 +976,44 @@ describe('end-to-end', function() {
...
+            .get('/url-does-not-exist')
+            .end(function(err, res) {
+              expect(res.error.text.match(/url-does-not-exist/));
+              done(err);
+          });
+        });
+      });
+    });
+
+    it('shows error behavior responses in production mode', function(done) {
```
bajtos: The test name should describe what is the expected behaviour/outcome. E.g. "it omits sensitive error details in production mode".

test/end-to-end.js
bajtos: The most important thing to check is that the stack trace is included in development mode.

bajtos: Ah, is `res.error.text` something coming from supertest? I was expecting to see a check verifying result JSON body.

I am also not sure what happens when the request fails in such way that there is no `res` object to return, how does supertest handle that case. To be safe, I usually do this:

```js
request(app)
  .get('/url-does-not-exist')
  .expect(404)
  .end(function(err, res) {
    if (err) return done(err);
    // verify & done
  });
```

test/end-to-end.js
```diff
@@ -970,6 +976,44 @@ describe('end-to-end', function() {
...
+        appRootDir: SANDBOX,
+        env: 'production'
+      };
+      boot(app, bootOptions, function(err) {
+        if (err) return done(err);
+        app.listen(0, function() {
+          request(app)
+            .get('/url-does-not-exist')
+            .end(function(err, res) {
+              expect(res.error.text.match(/url-does-not-exist/));
```
bajtos: In production mode, we need to check that the error response does not contain the stack trace. To make the assertion future-proof, we should not rely on particular (deep) property name like `error.stack`, but scan the string response body instead.

Also handle `err` from supertest as mentioned in https://github.com/strongloop/loopback-workspace/pull/285/files#r66961915

loay:
@bajtos PTAL

loay:
@slnode test please

loay:
@bajtos PTAL again

templates/projects/empty-server/data.js
```diff
@@ -79,7 +79,8 @@ template.server = {
       },
       cors: false,
       errorHandler: {
-        disableStackTrace: false,
+        debug: true,
```
bajtos: `debug` should be set to `false` by default in scaffolded projects.

bajtos: On the second thought, I think it may be best to disable the built-in error handler in strong-remoting and always use strong-error-handler instead.

```diff
   cors: false
-  errorHandler: {
-    disableStackTrace: false
- },
+  handleErrors: false,
 }},
```

That way we don't have to duplicate the configuration of strong-error-handler in multiple places.

test/end-to-end.js
```diff
@@ -970,6 +975,47 @@ describe('end-to-end', function() {
...
+        appRootDir: SANDBOX,
+        env: 'development'
+      };
+      boot(app, bootOptions, function(err) {
+        if (err) return done(err);
+        app.listen(0, function() {
+          request(app)
+            .get('/url-does-not-exist')
+            .expect(404)
+            .end(function(err, res) {
```
bajtos: unhandled `err`

test/end-to-end.js
```diff
@@ -970,6 +975,47 @@ describe('end-to-end', function() {
...
+        appRootDir: SANDBOX,
+        env: 'production'
+      };
+      boot(app, bootOptions, function(err) {
+        if (err) return done(err);
+        app.listen(0, function() {
+          request(app)
+            .get('/url-does-not-exist')
+            .expect(404)
+            .end(function(err, res) {
```
bajtos: unhandled err

test/end-to-end.js
```diff
@@ -970,6 +975,47 @@ describe('end-to-end', function() {
...
+              expect(JSON.stringify(res.body, null, 2)).to.not.contain(__filename)
```
bajtos: Please add a comment explaining what is the intent of this assertion, e.g.

```js
// Assert that the response body does not contain stack trace.
// We want the assertion to be robust and keep working even
// if the property name storing stack trace changes in the future,
// therefore we test full response body.
```

Also I just realised that `_filename` most likely won't be in the stack trace, because the error is created deep inside strong-remoting. Please check what is the actual `res.body` when `debug:true` and use a different string to detect stack trace. (In other words, make sure to see this test fail, to ensure the assertion is written in a way that can fail.)

bajtos:
@loay sorry for the back and forth, I realised my advise about checking the missing stack trace was wrong :(

templates/projects/empty-server/data.js
loay: how about sever/config.js ?
```
    "errorHandler": {
      "debug": true,
      "log": true
    }
```

loay:
@bajtos 
Now worries. It is just that `res.body` is giving an empty object, so I think that's something I need to figure out. Also, res.error works fine and it is giving stack in both dev and production !

loay:
@bajtos PTAL

test/end-to-end.js
```diff
@@ -970,6 +977,58 @@ describe('end-to-end', function() {
...
+        boot(app, bootOptions, function(err) {
+          if (err) return done(err);
+          app.listen(0, function() {
+            request(app)
+              .get('/url-does-not-exist')
+              .expect(404)
+              .end(function(err, res) {
+                app.use(function(err, res) {
+                  var responseBody = JSON.stringify(res.body);
+                  console.log(responseBody);
```
bajtos: Forgotten `console.log`. If you find it useful to have a log around, then please use `debug` instead.

In case `debug` is not defined in this file yet: in tests, we usually define `debug` at the top of the test file this way:

```js
var debug = require('debug')('test');
```

test/end-to-end.js
```diff
@@ -970,6 +977,58 @@ describe('end-to-end', function() {
...
+          env: 'development'
+        };
+        boot(app, bootOptions, function(err) {
+          if (err) return done(err);
+          app.listen(0, function() {
+            request(app)
+              .get('/url-does-not-exist')
+              .expect(404)
+              .end(function(err, res) {
+                app.use(function(err, res) {
```
bajtos: This doesn't work, the handler is never called as can be seen from no `console.log(responseBody)` appearing in the output of `mocha`.

test/end-to-end.js
```diff
@@ -970,6 +977,58 @@ describe('end-to-end', function() {
...
+    describe('error handling in dev/prod', function() {
+      it('includes sensitive error details in development mode', function(done) {
+        var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+        var bootOptions = {
+          appRootDir: SANDBOX,
+          env: 'development'
+        };
+        boot(app, bootOptions, function(err) {
+          if (err) return done(err);
+          app.listen(0, function() {
```
bajtos: No need to call `app.listen(0)`, `request(app)` takes care of that automatically.

bajtos:
> Now worries. It is just that res.body is giving an empty object, so I think that's something I need to figure out. Also, res.error works fine and it is giving stack in both dev and production !

From what I can tell, it looks like the application is misconfigured, strong-error-handler is not installed and the response contains empty body because it's produced by express default error handler.

Debug log:

```
  (...)
  superagent get http://127.0.0.1:60993/url-does-not-exist +22ms
  express:router dispatching GET /url-does-not-exist +20ms
  express:router query  : /url-does-not-exist +3ms
  express:router expressInit  : /url-does-not-exist +1ms
  finalhandler default 404 +3ms
  superagent get http://127.0.0.1:60993/url-does-not-exist -> 404 +12ms
  superagent end get http://127.0.0.1:60993/url-does-not-exist +6ms
```

I don't think there should be `finalhandler` entry when the request is handled by a middleware (like strong-error-handler).

Additionally, if you look at `res.error`, it contains the stack trace from the client side, not the server error.

```
   { Error: cannot GET /url-does-not-exist (404)
       at Response.toError (/Users/bajtos/src/loopback/workspace/node_modules/superagent/lib/node/response.js:106:13)
       at Response.setStatusProperties (/Users/bajtos/src/loopback/workspace/node_modules/superagent/lib/node/response.js:183:12)
       at new Response (/Users/bajtos/src/loopback/workspace/node_modules/superagent/lib/node/response.js:42:8)
       at IncomingMessage.<anonymous> (/Users/bajtos/src/loopback/workspace/node_modules/superagent/lib/node/index.js:918:22)
       at emitNone (events.js:91:20)
       at IncomingMessage.emit (events.js:185:7)
       at endReadableNT (_stream_readable.js:926:12)
       at _combinedTickCallback (internal/process/next_tick.js:74:11)
       at process._tickDomainCallback (internal/process/next_tick.js:122:9)
       at process.fallback (/Users/bajtos/src/loopback/workspace/node_modules/async-listener/index.js:450:15)
     status: 404,
     text: 'Cannot GET /url-does-not-exist\n',
     method: 'GET',
     path: '/url-does-not-exist' },
```

@loay Since this change is not urgent, I am going to use this as a learning opportunity for you. Here is what I'd like you to do:

1) Run only the first test from the new tests you are adding, for example by running 

```
$ mocha -g 'dev/prod.*develo'`
```

(There are other ways, pick whatever works best for you. Just make sure to include only tests that use the same sandbox as the test you are writing.)

2) After mocha finishes, you will have the scaffolded application left in `test/sandbox` directory. You can run `node .` there to start the app and call its REST API directly.

Right now, it fails with the following error:

```
$ node .
module.js:561
    throw err;
    ^

SyntaxError: /Users/bajtos/src/loopback/workspace/test/sandbox/server/middleware.development.json: Unexpected end of JSON input
```

Do not fix this error yet!

3) This has discovered a problem in your test - how comes the malformed JSON did not fail the test with? Find out the answer and fix this problem first, before making any changes in the templates. You should fix the problem in all your new tests that are affected.

4) Now it's time to fix that JSON template file (you are missing a closing curly brace).

5) After you get your tests to pass with the new implementation in place, make sure they also fail when templates change (a regression occurs). For example, you can temporarily switch `debug` to the opposite value (`debug:false` in dev, `debug:true` in production), and verify that the tests are failing with a descriptive error message pointing out to this problem.

loay:
@bajtos PTAL. Thanks.

PR #286, opened 2016-06-15
Add MQ Light to available connectors

qpresley:
@bajtos, I'm adding the new MQ Light connector to available-connectors and also fixing an issue from a previous PR which resulted in duplicate entries for dashDB


bajtos:
LGTM, please squash the commits before merging.

PR #290, opened 2016-06-22
Add supportedLBVersion to metadata

jannyHou:
Connect to strongloop-internal/scrum-loopback#926

Related proposal: https://github.com/strongloop/generator-loopback/pull/207#issuecomment-227672375

In this pr:
[1]. I added a function called `getAvailableLBVersion` to return the available versions. Now it's hard-coded. Not sure is it worth to get info from npm or store the available versions in a file, like how we store available connectors.

[2]. I added metadata `supportedLBVersions` to each template.
There is a flag in the array because of the custom merge function:
https://github.com/strongloop/loopback-workspace/blob/master/common/models/workspace.js#L143
Array with flag `skipInherits` will ignore the value from parent template when merge.
And remove the flag when generating returned available templates.
Test cases added.

### DIFF
--- a/common/models/workspace.js
+++ b/common/models/workspace.js
@@ -50,6 +50,22 @@ module.exports = function(Workspace) {
      */
 
     /**
+     * Get list of available loopback Versions.
+     *
+     * @callback {Function} callback
+     * @param {Error} err
+     * @param {Object} availableLBVersions
+     */
+
+    Workspace.getAvailableLBVersions = function(cb) {
+      var availableLBVersions = {
+        '2.x': { description: 'stable' },
+        '3.x': { description: 'pre-release, alpha quality' },
+      };
+      cb(null, availableLBVersions);
+    };
+
+    /**
      * Get an array of available template names.
      *
      * @callback {Function} callback
@@ -90,6 +106,7 @@ module.exports = function(Workspace) {
           return {
             name: name,
             description: data.description,
+            supportedLBVersions: data.supportedLBVersions,
           };
         });
         cb(null, templates);
@@ -129,6 +146,7 @@ module.exports = function(Workspace) {
         var t = template.inherits[ix];
         var data = this._loadProjectTemplate(t);
         if (!data) return null; // the error was already reported
+        delete data.supportedLBVersions;
         sources.unshift(data);
       }
       /* eslint-enable one-var */
--- a/templates/projects/api-server/data.js
+++ b/templates/projects/api-server/data.js
@@ -6,6 +6,8 @@ var template = module.exports;
 
 template.description = 'A LoopBack API server with local User auth';
 
+template.supportedLBVersions = ['2.x', '3.x'];
+
 template.inherits = [
   'empty-server',
 ];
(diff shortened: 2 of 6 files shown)


common/models/workspace.js
```diff
@@ -90,6 +90,7 @@ module.exports = function(Workspace) {
           return {
             name: name,
             description: data.description,
+            supportedLBVersion: data.supportedLBVersion,
```
bajtos: Please use plural ("...Versions") in the name.

bajtos:
@jannyHou please add a unit-test to verify that the new metadata is correctly copied to the method output and check why CI builds are failing.

common/models/workspace.js
jannyHou: Sure. got it.

jannyHou: CI fails due to the deep equal assert of template metadata. Will update those test cases too.

jannyHou:
@slnode test please

jannyHou:
@bajtos Please review again. Thank you. I add some details at the beginning to explain my change:
https://github.com/strongloop/loopback-workspace/pull/290#issue-161793002

common/models/workspace.js
```diff
@@ -142,7 +163,12 @@ module.exports = function(Workspace) {
       // when merging arrays, concatenate them (lodash replaces by default)
       sources.push(function templateMergeCustomizer(a, b) {
         if (_.isArray(a)) {
-          return a.concat(b);
+          // Skip the inherit of supportedLBVersions
+          if (a.indexOf('skipInherits') > -1) {
```
bajtos: I find the "skipInherits" flag not very elegant, I am proposing a different solution: remove "supportedLBVersions" property before we add the inherited template data to the list of sources, see lines 126-134 (147-155) above.

```js
       var sources = [template];
       /* eslint-disable one-var */
       if (template.inherits) for (var ix in template.inherits) {
         var t = template.inherits[ix];
         var data = this._loadProjectTemplate(t);
         if (!data) return null; // the error was already reported
         delete data.supportedLBVersions; //<< added by me
```

Thoughts?
         sources.unshift(data);
       }
       /* eslint-enable one-var */
```

Comment (jannyHou) [inline common/models/workspace.js]: lol much better than mine. Thank you.

**Comment by jannyHou** [comment]
@bajtos Applied change. Please review again :)

File: common/models/workspace.js [hunk 3]
```diff
@@ -50,6 +50,20 @@ module.exports = function(Workspace) {
...
+     * @return {Object} availableLBVersions
+     */
+
+    Workspace.getAvailableLBVersions = function() {
+      var availableLBVersions = {
+        '2.x': { description: 'stable' },
+        '3.x': { description: 'pre-release, alpha quality' },
+      };
+      return availableLBVersions;
+    };
```
bajtos: We need to make this method async, so that it can be exposed via REST API for consumption by APIC Editor (no need to add remoting metadata right now though).

```js
Workspace.getAvailableLBVersions = function(cb) {
  // ...
  cb(null, availableLBVersions);
};
```

See e.g. `getAvailableTemplates` which is already async.

bajtos:
One last comment, the rest LGTM. No further reviews are needed.

PR #291, opened 2016-06-23
Add acl config to new templates

jannyHou:
Fix

### DIFF
--- a/templates/projects/hello-world/data.js
+++ b/templates/projects/hello-world/data.js
@@ -22,10 +22,39 @@ template.server = {
 
   modelConfigs: [
     {
+      name: 'User',
+      dataSource: 'db',
+    },
+    {
+      name: 'AccessToken',
+      dataSource: 'db',
+      public: false,
+    },
+    {
+      name: 'ACL',
+      dataSource: 'db',
+      public: false,
+    },
+    {
+      name: 'RoleMapping',
+      dataSource: 'db',
+      public: false,
+    },
+    {
+      name: 'Role',
+      dataSource: 'db',
+      public: false,
+    },
+    {
       name: 'Message',
       dataSource: null,
     },
   ],
 
-  datasources: [],
+  datasources: [ 
+    {
+      name: 'db',
+      connector: 'memory',
+    },
+  ],
 };
--- a/templates/projects/hello-world/files/server/boot/authentication.js
+++ b/templates/projects/hello-world/files/server/boot/authentication.js
@@ -1,0 +1,4 @@
+module.exports = function enableAuthentication(server) {
+  // enable authentication
+  server.enableAuth();
+};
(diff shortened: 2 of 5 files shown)


jannyHou:
@bajtos @ritch @raymondfeng Please review. Thanks.

And if `server.enableAuth();` is necessary for an app, shall we move it to server/server.js instead of relying on boot script?

bajtos:
> And if `server.enableAuth();` is necessary for an app, shall we move it to server/server.js instead of relying on the boot script to run it?

@jannyHou IIRC we have intentionally placed that code in a boot script, to make it easy to remove it if users don't want auth in their app. Another idea was to create apps without auth first and then create a generator for enabling authentication - this generator then can add that boot script. In both cases, manipulating javascript code in server/server.js would be much more difficult.

templates/projects/hello-world/data.js
```diff
@@ -22,10 +22,39 @@ template.server = {
...
     },
   ],
 
-  datasources: [],
+  datasources: [ 
+    {
+      name: 'db',
+      connector: 'memory',
+    },
+  ],
```
bajtos: I believe @ritch explicitly wanted an example where there are no datasources. @ritch are you ok with modifying this "hello world" template to include in-memory datasource and the whole ACL setup?

If we go down this route, then I think we should also extend the hello-world endpoint to require some authentication, e.g. allow only authorized users to invoke the method.

Thoughts?

Last but not least, I'd like to see unit-tests demonstrating ACLs in hello-world app.

bajtos: Ah, github diff did not make it clear to me that there alredy is a unit-test for ALCs in hello world.

bajtos:
The code change looks mostly good, but I am not entirely sure whether this is the direction we want to take.

Since the original problem was that @cfjedimaster was not aware that his scaffolded app did not have authentication enabled, perhaps we should fix LoopBack runtime instead to report a warning when there are models with ACLs defined, but authentication was not enabled at app level.

Thoughts?

cfjedimaster:
So dumb question - are you saying ACLs do nothing if server.enableAuth(); isn't run? That seems... odd. Why wouldn't it always be on, and if I don't want to use it, simply don't use ACLs? (And then I'd roll my own.)

raymondfeng:
LGTM. Let's fix the security gap for now and refine it with more ideas later.

bajtos:
: 

So dumb question - are you saying ACLs do nothing if server.enableAuth(); isn't run? That seems... odd. Why wouldn't it always be on, and if I don't want to use it, simply don't use ACLs? (And then I'd roll my own.)

See https://github.com/strongloop/loopback/issues/2466.

I don't remember why we decided to make auth optional, perhaps @raymondfeng or @ritch remember?

> LGTM. Let's fix the security gap for now and refine it with more ideas later.

Fair enough, let's land this patch as it is.

PR #292, opened 2016-06-23
Add strong-error-handler to project templates

loay:
connect to strongloop-internal/scrum-loopback#896
connect to https://github.com/strongloop/loopback-boot/issues/193
reference for previous discussion: https://github.com/strongloop/loopback-workspace/pull/285

### DIFF
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -981,6 +981,56 @@ describe('end-to-end', function() {
         .end(done);
     }
   });
+
+  it('includes sensitive error details in development mode', function(done) {
+    var loopback = require(SANDBOX + '/node_modules/loopback');
+    var boot = require(SANDBOX + '/node_modules/loopback-boot');
+    var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+    var bootOptions = {
+      appRootDir: SANDBOX + '/server',
+      env: 'development'
+    };
+    boot(app, bootOptions, function(err) {
+      if (err) return done(err);
+      request(app)
+        .get('/url-does-not-exist')
+        .expect(404)
+        .end(function(err, res) {
+          if (err) return done (err);
+          var responseBody = JSON.stringify(res.body);
+          expect(responseBody).to.include('stack');
+
+          done();
+      });
+    });
+  });
+
+  it('omits sensitive error details in production mode', function(done) {
+    var loopback = require(SANDBOX + '/node_modules/loopback');
+    var boot = require(SANDBOX + '/node_modules/loopback-boot');
+    var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+    var bootOptions = {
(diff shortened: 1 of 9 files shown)


loay:
@bajtos PTAL

test/end-to-end.js
```diff
@@ -983,6 +983,56 @@ describe('end-to-end', function() {
   });
 });
 
+it('includes sensitive error details in development mode', function(done) {
```
bajtos: These two tests should live inside `describe('end-to-end')` block, since they are relying on SANDBOX being initialised with an app.

test/end-to-end.js
```diff
@@ -983,6 +983,56 @@ describe('end-to-end', function() {
...
+  var loopback = require(SANDBOX + '/node_modules/loopback');
+  var boot = require(SANDBOX + '/node_modules/loopback-boot');
+  var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+  var bootOptions = {
+    appRootDir: SANDBOX + '/server',
+    env: 'development'
+  };
+  boot(app, bootOptions, function(err) {
+    if (err) return done(err);
+      request(app)
```
bajtos: please remove the extra level of indentation.

test/end-to-end.js
```diff
@@ -983,6 +983,56 @@ describe('end-to-end', function() {
...
+        .expect(404)
+        .end(function(err, res) {
+          // Assert that the response body does not contain stack trace.
+          // We want the assertion to be robust and keep working even
+          // if the property name storing stack trace changes in the future,
+          // therefore we test full response body.
+           if (err) return done(err);
+           var responseBody = JSON.stringify(res.body);
+           expect(responseBody).to.not.include('stack');
+           expect(responseBody).to.not.contain('stack');
```
bajtos: Hmm, this looks suspicious to me.

http://chaijs.com/api/bdd/#method_include
> The include and contain assertions can be used as either property based language chains or as methods to assert the inclusion of an object in an array or a substring in a string. When used as language chains, they toggle the contains flag for the keys assertion.

IIUC, `include` and `contain` are aliases for the same assertion, there is no need to run both of them.

test/end-to-end.js
```diff
@@ -983,6 +983,56 @@ describe('end-to-end', function() {
...
+  boot(app, bootOptions, function(err) {
+    if (err) return done(err);
+      request(app)
+        .get('/url-does-not-exist')
+        .expect(404)
+        .end(function(err, res) {
+          // Assert that the response body does not contain stack trace.
+          // We want the assertion to be robust and keep working even
+          // if the property name storing stack trace changes in the future,
+          // therefore we test full response body.
```
bajtos: Indentation is off, please fix.

bajtos:
LGTM

PR #294, opened 2016-06-24
update strong-error-handler test cases

loay:

PR #298, opened 2016-06-28
Fix tests failing on CI

bajtos:
Depends on https://github.com/strongloop/loopback-boot/pull/194 and supersedes https://github.com/strongloop/loopback-workspace/pull/294

### DIFF
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -152,6 +152,56 @@ describe('end-to-end', function() {
         .expect('Content-Type', /json/)
         .end(done);
     });
+
+    it('includes sensitive error details in development mode', function(done) {
+      var loopback = require(SANDBOX + '/node_modules/loopback');
+      var boot = require(SANDBOX + '/node_modules/loopback-boot');
+      var app = loopback({ localRegistry: true, loadBuiltinModels: true });
+      var bootOptions = {
+        appRootDir: SANDBOX + '/server',
+        env: 'development',
+      };
+      boot(app, bootOptions, function(err) {
+        if (err) return done(err);
+        request(app)
+          .get('/url-does-not-exist')
+          .expect(404)
+          .end(function(err, res) {
+            if (err) return done (err);
+            var responseBody = JSON.stringify(res.body);
+            expect(responseBody).to.include('stack');
+
+            done();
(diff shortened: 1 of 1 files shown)


bajtos:
To explain: this change is moving the two recently added test from top-level "end-to-end" block inside nested "empty-server template" block, where the sandbox app required by these two test cases is properly scaffolded by before/beforeEach hooks.

/to @jannyHou @0candy please review
/cc @richardpringle

0candy:
The changes LGTM as you're just moving the test cases. However, all of CI is failing so it doesn't look like it is fixing the issue.

bajtos:
@slnode test please

bajtos:
I have published `loopback-boot@2.20.0` to npmjs.org, that seems to have fixed the problem on my machine.

My guess is that `npm install` run from within the test suite uses the public npm registry, not the staging. @rmg is it something we can fix in our CI build tool? I would expect `npm install` executed by a test to use the same registry as `npm install` used to install the dependencies before running the tests.

bajtos:
@slnode test please

bajtos:
@0candy @jannyHou the remaining test failures seems to be intermittent. For example, cis-jenkins tests passed on all platforms, while pr-builder failed on Node v6 with errors that seem unrelated:

> ```
Test Result (5 failures / -3)
end-to-end autoupdate.end-to-end autoupdate before
end-to-end discovery.end-to-end discovery before
end-to-end testConnection.end-to-end testConnection before
end-to-end start/stop/restart.end-to-end start/stop/restart before
Middleware Middleware.create(def, cb).should be able to create multiple entries
```

(see http://ci.strongloop.com/job/loopback-workspace/label=amazon-6/4378/)

I am going to re-run CI again to see if we manage to get a green build. Even if not, I am proposing to merge this pull request, as it's a big step towards getting CI green.

bajtos:
@slnode test please

jannyHou:
@bajtos LGTM and I think we can merge it now. The failing ones are downstream for apic and apim.
`loopback-workspace` also fails due to apic/apim failures.

bajtos:
Landed, thank you @0candy and @0candy for the review.

Issue #302, opened 2016-07-16
Calling Workspace.createFromTemplate via JS api 2 times results in dup-model error


kraman:
When calling https://github.com/strongloop/loopback-workspace/blob/master/common/models/workspace.js#L363 with the notes template via JS API,
it seems like some model information is being cached. We get an error 

`The 'ModelConfig' instance is not valid. 'name' is not unique (value: "User").`

chandadharap:
Assigned to Candy. We need this fixed for code freeze on Monday 7/18/16. Affect Multi-project UX 5030 feature. @jtary and @kraman FYI.

Reference:
https://github.ibm.com/apimesh/scrum-platform-foundation/issues/1257

bajtos:
@0candy 

> I am trying to reset the workspace after createFromTemplate but I don’t think that’s the right way to fix it. (edited)
> Should this be supported on the workspace side or should workspace be reset in apic after every loopback project is created?

IMO, we should fix `loadWorkspace` (added by https://github.com/strongloop/loopback-workspace/pull/240) to to reset the internal state when WORKSPACE_DIR is changed.

I think the implementation we use in tests, where we delete all workspace models instances before changing the project directory, may not work here - I would expect that such delete would trigger save operation and thus the project files will be emptied as a result. It's not a problem in tests, because `resetWorkspace` is usually called after all files from the project dir were removed.

It is crucial to add a unit-test reproducing the current issue, to ensure we understand correctly the problem and to prevent regressions in the future.

Things to watch out for: after a new project is loaded,
 - The old project should remain intact, i.e. no project files are changed as the result of switch.
 - The new project is not touched, i.e. no project files are changes as the result of switch. This does not matter when creating a new project, but it's important when switching between two existing projects.

bajtos:
Once `loadWorkspace` is fixed, we should rework unit-test to use `loadWorkspace` instead of test-provided `resetWorkspace`. That may help us to uncover further edge cases to fix. However, I think it's ok to leave that as the second step, out of scope of the urgent fix.

PR #303, opened 2016-07-17
Fix createFromTemplate duplicate model error

0candy:
Connect to https://github.com/strongloop/loopback-workspace/issues/302

### DIFF
--- a/test/support.js
+++ b/test/support.js
@@ -19,6 +19,10 @@ expectFileExists = function(file) {
   assert(fs.existsSync(file), file + ' does not exist');
 };
 
+expectFileNotExists = function(file) {
+  assert(!fs.existsSync(file), file + ' does exist');
+};
+
 getPath = function(relativePath) {
   return ConfigFile.toAbsolutePath(relativePath);
 };
--- a/test/workspace.js
+++ b/test/workspace.js
@@ -249,4 +249,61 @@ describe('Workspace', function() {
       });
     });
   });
+
+  describe('Creating 2 projects from template', function() {
+    beforeEach(givenEmptySandbox);
+    afterEach(function setWorkspaceToSandboxDir() {
+      process.env.WORKSPACE_DIR = SANDBOX;
+    });
+
+    it('Create note app & hello world app in same directory and switching workspace',
+      function(done) {
+        async.series([
+          function(done) {
+            fs.mkdir(SANDBOX + '/noteapp', done);
+          },
+          function(done) {
+            Workspace.loadWorkspace(SANDBOX + '/noteapp', function() {
+              expect(process.env.WORKSPACE_DIR).to.equal(SANDBOX + '/noteapp');
+              done();
+            });
+          },
+          function(done) {
+            app.models.Workspace.createFromTemplate('notes', 'noteapp', {},
+              function(err) {
+                if (err) return done(err);
+                expectFileExists(getPath('server/server.js'));
+                expectFileExists(getPath('server/boot/authentication.js'));
+                expectFileExists(getPath('common/models/note.js'));
+                expectFileExists(getPath('common/models/note.json'));
+                done();
+              }
+            );
+          },
+          function(done) {
+            fs.mkdir(SANDBOX + '/helloworldapp', done);
+          },
+          function(done) {
+            Workspace.loadWorkspace(SANDBOX + '/helloworldapp', function() {
+              expect(process.env.WORKSPACE_DIR).to.equal(SANDBOX + '/helloworldapp');
+              done();
+            });
+          },
+          function(done) {
+            app.models.Workspace.createFromTemplate('hello-world', 'helloworldapp', {},
+              function(err) {
+                if (err) return done(err);
+                expectFileExists(getPath('server/server.js'));
+                expectFileExists(getPath('server/boot/authentication.js'));
+                expectFileExists(getPath('common/models/message.js'));
+                expectFileExists(getPath('common/models/message.json'));
+                expectFileNotExists(getPath('common/models/note.js'));
+                expectFileNotExists(getPath('common/models/note.json'));
+                done();
+              }
+            );
+          },
+        ], done);
+      });
+  });
 });
--- a/common/models/workspace.js
+++ b/common/models/workspace.js
@@ -651,9 +651,11 @@ module.exports = function(Workspace) {
     });
 
     Workspace.loadWorkspace = function(path, cb) {
-      process.env.WORKSPACE_DIR = path;
-      debug(process.env.WORKSPACE_DIR);
-      setImmediate(cb);
+      app.dataSources.db.connector.saveToFile(null, function() {
+        process.env.WORKSPACE_DIR = path;
+        debug(process.env.WORKSPACE_DIR);
+        app.dataSources.db.connector.loadFromFile(cb);
+      });
     };
 
     loopback.remoteMethod(Workspace.loadWorkspace, {


0candy:
@bajtos PTAL

0candy:
@raymondfeng @ritch @jannyHou Could someone please review the changes. Thanks!

raymondfeng:
@0candy Your change looks good to me. @ritch should be the final approver. 

Can you ask @kraman to verify your changes with his setup?

0candy:
@kraman Can you please verify that this change will fix the issue with your setup? Thanks!

test/support.js
```diff
@@ -19,6 +19,10 @@ expectFileExists = function(file) {
   assert(fs.existsSync(file), file + ' does not exist');
 };
 
+expectFileNotExists = function(file) {
```
superkhau: `expectFileToNotExist`

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
...
+    beforeEach(givenEmptySandbox);
+    afterEach(function setWorkspaceToSandboxDir() {
+      process.env.WORKSPACE_DIR = SANDBOX;
+    });
+
+    it('Create note app & hello world app in same directory and switching workspace',
+      function(done) {
+        async.series([
+          function(done) {
+            fs.mkdir(SANDBOX + '/noteapp', done);
```
superkhau: You could probably extract this function to support:

```
createSandboxDir = function(dirName) {
  ...
};
```

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
...
+                expectFileExists(getPath('server/boot/authentication.js'));
+                expectFileExists(getPath('common/models/note.js'));
+                expectFileExists(getPath('common/models/note.json'));
+                done();
+              }
+            );
+          },
+          function(done) {
+            fs.mkdir(SANDBOX + '/helloworldapp', done);
+          },
```
superkhau: ditto

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
...
+
+  describe('Creating 2 projects from template', function() {
+    beforeEach(givenEmptySandbox);
+    afterEach(function setWorkspaceToSandboxDir() {
+      process.env.WORKSPACE_DIR = SANDBOX;
+    });
+
+    it('Create note app & hello world app in same directory and switching workspace',
+      function(done) {
+        async.series([
```
superkhau: You could clean this up by taking out all the inline functions and extracting some to the bottom of the file and some to `support.js`. Once that's done, you can write:

```
async.series([
  createSandboxDir.bind('/noteapp'),
  loadWorkspace.bind('/noteapp'),
  testFilesDoNotExist.bind('/noteapp'),
  createSandboxDir.bind('/helloworldapp'),
  ...
]);
```

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
       });
     });
   });
+
+  describe('Creating 2 projects from template', function() {
+    beforeEach(givenEmptySandbox);
+    afterEach(function setWorkspaceToSandboxDir() {
+      process.env.WORKSPACE_DIR = SANDBOX;
+    });
```
superkhau: move func to bottom of the file outside the describe block, then u can:

```
beforeEach(givenEmptySandbox);
afterEach(setWorkspaceToSandboxDir);
```

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
...
+
+    it('Create note app & hello world app in same directory and switching workspace',
```
superkhau: it('creates `noteapp` & `helloworldapp` in the same dir while switching workspaces', ...

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
       });
     });
   });
+
+  describe('Creating 2 projects from template', function() {
```
superkhau: `describe('Workspace.createFromTemplate(...)', function(...)` like the other descriptions above.

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
...
+            Workspace.loadWorkspace(SANDBOX + '/helloworldapp', function() {
+              expect(process.env.WORKSPACE_DIR).to.equal(SANDBOX + '/helloworldapp');
+              done();
+            });
+          },
+          function(done) {
+            app.models.Workspace.createFromTemplate('hello-world', 'helloworldapp', {},
+              function(err) {
+                if (err) return done(err);
+                expectFileExists(getPath('server/server.js'));
```
superkhau: expectFileToExist

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
...
+                expectFileExists(getPath('server/boot/authentication.js'));
```
superkhau: expectFileToNotExist

test/support.js
superkhau: nitpick: file = filePath

test/support.js
```diff
@@ -19,6 +19,10 @@ expectFileExists = function(file) {
...
+  assert(!fs.existsSync(file), file + ' does exist');
```
superkhau: existsSync is deprecated in new node. Should use fs.statSync or fs.accessSync instead: see https://nodejs.org/api/fs.html#fs_fs_existssync_path

superkhau:
@0candy Reviewed. Logic is LGTM, other things are just formatting/convention/nitpicks. I assume this one needs to go out fast. ;)

test/workspace.js
```diff
@@ -249,4 +249,61 @@ describe('Workspace', function() {
       });
     });
   });
+
+  describe('Creating 2 projects from template', function() {
+    beforeEach(givenEmptySandbox);
```
jannyHou: @0candy Do we need to `beforeEach(resetWorkspace)` ?

0candy: No, we don't need to `resetWorkspace` as I'm using `loadWorkspace`. Eventually, @bajtos suggest changing all the `resetWorkspace` to use `loadWorkspace`.

test/workspace.js
0candy: Did you mean to move the test cases into the previous describe block above?
https://github.com/strongloop/loopback-workspace/blob/master/test/workspace.js#L100

kraman:
@anthonyettinger verified that this fixes the issue. Thanks @0candy

test/support.js
bajtos: `fs.accessSync` is not available on Node v0.10.

test/workspace.js
bajtos: This is subjective, I personally prefer the current style.

test/support.js
bajtos: Since we already have `expectFileExists` (see few lines higher up in this file), I think it's better to keep `expectFileNotExists` for consistency.

bajtos:
@0candy I have reviewed the patch too, it LGTM.

I am proposing to land & release it as is right now. Stylistic issues pointed out by @superkhau can be addressed in a follow-up pull request IMO.

test/workspace.js
jannyHou: Thanks :) LGTM now

test/workspace.js
superkhau: No i mean the other describe blocks use another naming convention:

Example:

```
describe('Workspace.loadWorkspace(dir, cb)', function() {
```

ritch:
We should work out a way to load isolated `Workspace`s. The code that worries me is the global state in `process.env.WORKSPACE_DIR`. I think we have something working (verified by @anthonyettinger for now), but am worried that this will lead to bugs in the future.

Something like:

```js
var workspaceA = new Workspace(pathToWorkspaceA);
var workspaceB = new Workspace(pathToWorkspaceB);

workspaceA.listProjects(); // => lists projects for this workspace
workspaceB.listProjects(); // => lists projects for this workspace
```

The problem is that we cannot do this with the current REST API. It just isn't possible with loopback... @0candy can you create an issue to track this? Then this can be landed and we can prioritize that fix.

chandadharap:
and best to link to that issue here @0candy

0candy:
Created issue for @ritch 's suggestion:
https://github.com/strongloop/loopback-workspace/issues/304

PR #305, opened 2016-07-19
Fix test descriptions and refactor

0candy:
Fix coding style comments from this PR: https://github.com/strongloop/loopback-workspace/pull/303

### DIFF
--- a/test/support.js
+++ b/test/support.js
@@ -15,12 +15,21 @@ var debug = require('debug')('workspace:test:support');
 
 global.Promise = require('bluebird');
 
+fileExistsSync = function(path) {
+  try {
+    fs.statSync(path);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
 expectFileExists = function(file) {
-  assert(fs.existsSync(file), file + ' does not exist');
+  assert(fileExistsSync(file), file + ' does not exist');
 };
 
 expectFileNotExists = function(file) {
-  assert(!fs.existsSync(file), file + ' does exist');
+  assert(!fileExistsSync(file), file + ' does exist');
 };
 
 getPath = function(relativePath) {
@@ -39,10 +48,14 @@ SANDBOX = path.resolve(__dirname, 'sandbox/');
 // tell the workspace to load files from the sandbox
 process.env.WORKSPACE_DIR = SANDBOX;
 
+createSandboxDir = function(dirName, cb) {
+  fs.mkdir(dirName, cb);
+};
+
 givenEmptySandbox = function(cb) {
   fs.remove(SANDBOX, function(err) {
     if (err) return cb(err);
-    fs.mkdir(SANDBOX, cb);
+    createSandboxDir(SANDBOX, cb);
   });
 
   // Remove any cached modules from SANDBOX
@@ -125,6 +138,11 @@ givenLB3Workspace = function(cb) {
     var options = { loopbackVersion: '3.x' };
     givenWorkspaceFromTemplate('empty-server', options, cb);
   });
+};
+
+setWorkspaceToSandboxDir = function() {
+  // tell the workspace to load files from the sandbox
+  process.env.WORKSPACE_DIR = SANDBOX;
 };
 
 function findOfType(name, type) {
--- a/test/workspace.js
+++ b/test/workspace.js
@@ -250,25 +250,23 @@ describe('Workspace', function() {
     });
   });
 
-  describe('Creating 2 projects from template', function() {
-    beforeEach(givenEmptySandbox);
-    afterEach(function setWorkspaceToSandboxDir() {
-      process.env.WORKSPACE_DIR = SANDBOX;
-    });
+  describe('Multi-project workspace support', function() {
+    before(givenEmptySandbox);
+    after(setWorkspaceToSandboxDir);
 
-    it('Create note app & hello world app in same directory and switching workspace',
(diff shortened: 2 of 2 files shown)


test/end-to-end.js
```diff
@@ -1043,6 +1043,63 @@ describe('end-to-end', function() {
         .end(done);
     }
   });
+
+  describe('Multi-project workspace support', function() {
+    beforeEach(givenEmptySandbox);
+    afterEach(function setWorkspaceToSandboxDir() {
+      process.env.WORKSPACE_DIR = SANDBOX;
+    });
```
superkhau: afterEach(setWorkspaceToSandboxDir);

Also subjective is beforeEach can be just "before" and afterEach can just be "after" since you only one test. The next person adding a test can then decide whether they want it for their test without having to modify the "Each" part (and add it if they deem it necessary).

test/end-to-end.js
```diff
@@ -1043,6 +1043,63 @@ describe('end-to-end', function() {
...
+                expectFileExists(getPath('common/models/message.js'));
+                expectFileExists(getPath('common/models/message.json'));
+                expectFileNotExists(getPath('common/models/note.js'));
+                expectFileNotExists(getPath('common/models/note.json'));
+                done();
+              }
+            );
+          },
+        ], done);
+      });
```
superkhau:
```
function setWorkspaceToSandboxDir() {
  process.env.WORKSPACE_DIR = SANDBOX;
}
```

test/end-to-end.js
```diff
@@ -1043,6 +1043,63 @@ describe('end-to-end', function() {
...
+  describe('Multi-project workspace support', function() {
+    beforeEach(givenEmptySandbox);
+    afterEach(function setWorkspaceToSandboxDir() {
+      process.env.WORKSPACE_DIR = SANDBOX;
+    });
+
+    it('switches workspace while creating two apps in the same dir',
+      function(done) {
+        async.series([
+          function(done) {
```
superkhau: maybe change all these internal dones to `next` and keep `done` as the main one on 1054 and 1100 for clarity.

superkhau:
Some nitpicks, but LGTM.

test/support.js
```diff
@@ -15,12 +15,21 @@ var debug = require('debug')('workspace:test:support');
 
 global.Promise = require('bluebird');
 
+fileExist = function(path) {
```
bajtos: Please rename to `fileExistsSync` to prevent confusion about whether this function expects a callback or returns the result synchronously.

test/end-to-end.js
```diff
@@ -1043,6 +1043,66 @@ describe('end-to-end', function() {
...
+                expectFileNotExists(getPath('common/models/note.js'));
+                expectFileNotExists(getPath('common/models/note.json'));
+                next();
+              }
+            );
+          },
+        ], done);
+      });
+
+    function setWorkspaceToSandboxDir() {
```
bajtos: I think this helper should be moved to `test/support.js`.

test/end-to-end.js
```diff
@@ -1043,6 +1043,66 @@ describe('end-to-end', function() {
         .end(done);
     }
   });
+
+  describe('Multi-project workspace support', function() {
```
bajtos: I am wondering what is the reason for moving this test from `test/workspace` to `test/end-to-end`?

The tests in `end-to-end` are running `npm install` and taking long time to complete, that's why they are separated in a standalone file.

AFAICT, your test only calls `createFromTemplate`, which is relatively fast and there are already other tests calling this method in `test/workspace`.

Can we move this test/describe block back to `test/workspace`?

superkhau: >I am wondering what is the reason for moving this test from test/workspace to test/end-to-end?

I asked her to move it in another code review as I felt it was not exactly a unit test. Ideally, we should create a new dir (ie. test/integration) for tests that are not unit based. IMO, there should be a `test/unit/workspace.js` with two sets of tests (one set for .loadWorkspace, a seperate unit for createFromTemplate). Then the test/integration can test them working together by calling both like this test is doing. e2e probably wasn't the best place, but at the time I felt it was better than mixing it in with the basic unit tests.

superkhau:
@bajtos Can you take over this review? I'm fine with whatever you suggest as a convention (ie. whether or not to create the new dirs now and split up the unit/integration or just to move it into test/workspace).

test/end-to-end.js
bajtos: I prefer consistency and have all workspace-related tests in the same file, that is `test/workspace.js`.

I agree we should eventually make more granular test files (`test/unit/workspace.js`, `test/integration/workspace.js`, etc.), but that's out of scope of this pull request.

Since `test/end-to-end` is not the right place for this new test, I strongly prefer to put it where other similar tests are already (`test/workspace.js`).

bajtos:
> Can you take over this review? I'm fine with whatever you suggest as a convention (ie. whether or not to create the new dirs now and split up the unit/integration or just to move it into test/workspace).

Sure.

@0candy please move the test back to `test/workspace` and this can be landed.

0candy:
@slnode test please

0candy:
@bajtos Fixed comments. PTAL

bajtos:
LGTM, :shipit:

bajtos:
@0candy actually, please squash the commits into a single one before landing ;)

PR #306, opened 2016-07-20
Fix crash due to missing optionalDependencies

jtary:
Fix issue were, under specific npm setups, workspace would crash on
startup do to optionalDependencies being undefined, and trying to
resolve the loopback version though it.

### DIFF
--- a/server/connector.js
+++ b/server/connector.js
@@ -200,9 +200,21 @@ connector._loadFromFile = function(cb) {
           if (err) return next(err);
           PackageDefinition.addToCache(cache, f.data);
           if (dir === '.') {
-            var loopBackVersion = f.data.dependencies['loopback'] ||
-              f.data.devDependencies['loopback'] ||
-              f.data.optionalDependencies['loopback'];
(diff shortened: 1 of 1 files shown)


jtary:
@raymondfeng @bajtos PTAL

server/connector.js
```diff
@@ -201,8 +201,12 @@ connector._loadFromFile = function(cb) {
           PackageDefinition.addToCache(cache, f.data);
           if (dir === '.') {
             var loopBackVersion = f.data.dependencies['loopback'] ||
-              f.data.devDependencies['loopback'] ||
-              f.data.optionalDependencies['loopback'];
+              f.data.devDependencies['loopback'];
```
raymondfeng: There will be similar situations where `dependencies` or `devDependencies` is undefined.

raymondfeng:
LGTM

jtary:
The CLI loopback failure is unrelated to this PR, as are the apm-ui failures. I'll take a look at fixing up the old apim-ui branches to prevent those failures in the future.

PR #307, opened 2016-07-26
workspace: remove "alpha" from 3.0 description

bajtos:
Address https://github.com/strongloop/generator-loopback/pull/205#issuecomment-230638249

@richardpringle @crandmck please review
cc @jannyHou


crandmck:
I think it's fine "as is," but if you want to be really clear, you could make it: 
```
'Pre-release; Do not use in production.' 
```

richardpringle:
Agree with @crandmck, either way is fine.

PR #308, opened 2016-07-29
Disable current-context in new projects

bajtos:
The current implementation of current context in LoopBack is not good enough for general use because it has too many limitations.

This commit changes the template to disable current-context middleware via `remoting.context` option.

See also strongloop/loopback#2559 and strongloop/loopback#2564

/to @raymondfeng @ritch please review
/cc @0candy @jannyHou @richardpringle


raymondfeng:
LGTM

PR #311, opened 2016-08-10
Fix CI timeouts and windows posix errors

0candy:
- Fix timeout errors
- Fix windows posix `test` command
- Fix mysql on windows by default table names are lower case.

Connect to https://github.com/strongloop/loopback-workspace/issues/322

### DIFF
--- a/bin/setup-mysql.js
+++ b/bin/setup-mysql.js
@@ -16,23 +16,30 @@ DATABASE = 'loopback_workspace_test';
 USER = 'lbws';
 PASSWORD = 'hbx42rec';
 
-var connection;
+var connection, password;
 async.series([
-  function setupConnection(next) {
+  function askForPassword(next) {
+    if (process.env.CI) {
+      password = '';
+      return next();
+    }
     read({
       prompt: g.f('Enter password for MySQL root user:'),
       silent: true,
-    }, function(err, password) {
+    }, function(err, pwd) {
       if (err) return next(err);
-
-      connection = mysql.createConnection({
-        user: 'root',
-        password: password,
-      });
-
-      g.log('Connecting');
-      connection.connect(next);
+      password = pwd;
+      next();
     });
+  },
+  function setupConnection(next) {
+    connection = mysql.createConnection({
+      user: 'root',
+      password: password,
+    });
+
+    g.log('Connecting');
+    connection.connect(next);
   },
   function createDatabase(next) {
     g.log('Creating database %s', DATABASE);
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "export-tests": true
   },
   "scripts": {
-    "pretest": "test -z \"$JENKINS_HOME\" || (echo | node bin/setup-mysql.js)",
+    "pretest": "node bin/setup-mysql.js",
     "lint": "eslint .",
     "test": "mocha",
     "posttest": "npm run lint"
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -30,7 +30,7 @@ var MYSQL_USER = 'lbws';
 var MYSQL_PASSWORD = 'hbx42rec';
 
 describe('end-to-end', function() {
-  this.timeout(10000);
+  this.timeout(15000);
 
   describe('empty-server template', function() {
     var app;
@@ -628,7 +628,7 @@ describe('end-to-end', function() {
         if (err) done(err);
         listTableNames(connection, function(err, tables) {
           if (err) done(err);
-          expect(tables).to.contain('CUSTOM');
+          expect(tables).to.match(/CUSTOM/i);
           done();
         });
       });
@@ -639,7 +639,9 @@ describe('end-to-end', function() {
         if (err) done(err);
         listTableNames(connection, function(err, tables) {
           if (err) done(err);
-          expect(tables).to.include.members(['CUSTOM', 'User', 'AccessToken']);
+          expect(tables).to.match(/CUSTOM/i);
+          expect(tables).to.match(/User/i);
+          expect(tables).to.match(/AccessToken/i);
           done();
         });
       });
@@ -647,7 +649,7 @@ describe('end-to-end', function() {
   });
 
   describe('discovery', function() {
-    this.timeout(10000);
+    this.timeout(15000);
 
     var connection;
     before(function(done) {
@@ -863,7 +865,7 @@ describe('end-to-end', function() {
 
     // The tests are forking new processes and setting up HTTP servers,
     // they requires more than 2 seconds to finish
-    this.timeout(10000);
+    this.timeout(20000);
 
     before(resetWorkspace);
     before(givenBasicWorkspace);
--- a/test/facet.js
+++ b/test/facet.js
@@ -8,6 +8,8 @@ var Facet = require('../').models.Facet;
 
 describe('Facet', function() {
   describe('Facet.create(def, cb)', function() {
+    this.timeout(15000);
+
     beforeEach(givenBasicWorkspace);
 
     it('should use name as the id', function(done) {
--- a/test/model-definition.js
+++ b/test/model-definition.js
@@ -13,6 +13,7 @@ var ModelRelation = app.models.ModelRelation;
 var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var ConfigFile = app.models.ConfigFile;
+var path = require('path');
 
 describe('ModelDefinition', function() {
   describe('CRUD', function() {
@@ -119,14 +120,15 @@ describe('ModelDefinition', function() {
   describe('ModelDefinition.getPath(app, obj)', function() {
     it('should return the configFile path if it exists', function() {
       var configFilePath = 'foo/bar/bat/baz.json';
-      var path = ModelDefinition.getPath('.', { name: 'MyModel',
+      var modelPath = ModelDefinition.getPath('.', { name: 'MyModel',
         configFile: configFilePath });
 
-      expect(path).to.equal(configFilePath);
+      expect(modelPath).to.equal(configFilePath);
     });
     it('should return construct configFile path', function() {
-      var path = ModelDefinition.getPath('.', { name: 'MyModel' });
-      expect(path).to.equal('models/my-model.json');
+      var configFilePath = 'models/my-model.json';
+      var modelPath = ModelDefinition.getPath('.', { name: 'MyModel' });
+      expect(modelPath).to.equal(path.normalize(configFilePath));
     });
   });
 
--- a/common/models/workspace.js
+++ b/common/models/workspace.js
@@ -520,7 +520,7 @@ module.exports = function(Workspace) {
         child.on('exit', function(code) {
           debug('child %s exited with code %s', child.pid, code);
           Workspace._child = null;
-          done(new Error(g.f('Child exited with code %s' + code)));
+          done(new Error(g.f('Child exited with code %s', code)));
         });
 
         // Wait until the child process starts listening


jannyHou:
I assume this pr is created to fix failings reported here :)
https://cis-jenkins.swg-devops.com/job/ds/job/loopback-workspace~master/62/console

My PR in generator-loopback has similar problem on all node 0.10 platforms:
One test case fails due to timeout, then others fail due to 
> Error: ENOTEMPTY, rmdir some_file_path

I thought it's a node 0.10 problem but seems the CI link I post above uses node 4.x.
Also I could not reproduce it on local with node 4.4.7, for example if I run two failing tests together but comment `done()` in the first one, the second one still pass, no ENOTEMPTY error.
I will try with 0.10.

0candy:
@jannyHou I seem to get different results each time. However, it passes on my local with node 0.10

jannyHou:
Link issue https://github.com/strongloop/loopback-workspace/pull/312

bajtos:
I am almost certainly sure that if workspace fails on Node v0.10 only, then it's caused by https://github.com/rxaviers/cldr-data-npm/pull/34

bajtos:
FWIW, rxaviers/cldr-data-npm#34 has been landed & released, I think this patch should not be needed any more. For example, CI build of https://github.com/strongloop/loopback-workspace/pull/312 is passing on Node v0.10 AFAICT.

@jannyHou @0candy

bajtos:
However, there seem to be other timing-related build failures on Node v0.12, see http://ci.strongloop.com/job/loopback-workspace/label=amazon-0.12/4757/consoleText

```

  1) end-to-end notes template "before all" hook:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
  

  2) end-to-end hello-world template "before all" hook:
     Error: EEXIST, mkdir '/mnt/jenkins/workspace/loopback-workspace/1de32925/test/sandbox'
      at Error (native)

  3) end-to-end start/stop/restart restarts the app:
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
      at null.<anonymous> (node_modules/loopback/node_modules/loopback-context/node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31)

  4) Facet Facet.create(def, cb) "before each" hook for "should use name as the id":
     Error: timeout of 10000ms exceeded. Ensure the done() callback is being called in this test.
      at null.<anonymous> (node_modules/loopback/node_modules/loopback-context/node_modules/continuation-local-storage/node_modules/async-listener/glue.js:188:31)
```

0candy:
@slnode test please

0candy:
I reran it and now it fails for node 6. I think there is still a timing issue not dependent on node version.

package.json
```diff
@@ -6,7 +6,7 @@
     "export-tests": true
   },
   "scripts": {
-    "pretest": "test -z \"$JENKINS_HOME\" || (echo | node bin/setup-mysql.js)",
```
rmg: The `echo |` is still needed because `node bin/setup-mysql.js` is prompting for a password.

superkhau: Will that work on Windows?

rmg: We can try it and find out. The alternative is to modify `bin/setup-mysql.js` so that it doesn't use `read()`.

0candy: `echo |` does not work for windows.

rmg:
Perhaps the `bin/setup-mysql.js` script should accept a `--auto` flag that makes it not prompt for a password and always exit successfully. If run _without_ the `--auto` flag, it could use the current behaviour of prompting for a password and exiting with failure if appropriate.

bajtos:
Let's step back and ask ourselves what is the `pretest` hook trying to achieve?

 - When the developer is running `npm test` locally, the SQL setup should be skipped. We assume they have already set up their database.
 - When running `npm test` on a CI server, we may have a fresh SQL server instance and need to setup the database & access credentials. The assumption is that the CI server has an admin user with an empty password.

In that light, I think we should change `bin/setup-mysql.js` to detect CI, skip `read` and use an empty password.

```js
function askForPassword(next) {
  if (process.env.CI) {
    password = '';
    return next();
  }
  read({
    prompt: g.f('Enter password for MySQL root user:'),
    silent: true,
  }, function(err, pwd) {
    if (err) return next(err);
    password = pwd;
    next();
  });
}
```

bajtos:
Moving here the discussion from https://github.com/strongloop/loopback-workspace/issues/322:

> If the script is relatively non-intrusive and developers don't mind it running every time they do npm test then it could be made unconditional.
> We should just remove the jenkins check altogether.

I disagree.

The script needs MySQL `admin` password to work correctly. Having to type the password every time I run `npm test` is rather intrusive to me personally.

superkhau:
>When running npm test on a CI server, we may have a fresh SQL server instance and need to setup the database & access credentials. The assumption is that the CI server has an admin user with an empty password.

Can the CI instance be set up to use MySQL default user (ie. root/nopass)? External users do not have access to CI anyways.

@bajtos Can we just write the script to assume root/nopass since that is how default MySQL instances come OOB anyways? Then leave it up to the user running `npm test` to set up their env however they want (whether it be CI or community, etc).

rmg:
> Can the CI instance be set up to use MySQL default user (ie. root/nopass)? External users do not have access to CI anyways.

@superkhau that _is_ how it is set up, and how this script is able to run. The problem is `bin/setup-mysql.js` is written in such a way that it essentially *requires* someone to "press return" instead of a non-interactive mechanism.

> When the developer is running `npm test` locally, the SQL setup should be skipped. We assume they have already set up their database.

I personally prefer modules to be a _little_ more turn-key and idempotent than that. Can we at least having `npm test` tell me (as a new contributor) what I need to do to get `npm test` to pass if I run it before running `bin/setup-mysql.js`?

> In that light, I think we should change bin/setup-mysql.js to detect CI, skip read and use an empty password.

FYI, the connectors use a different convention and instead check for and use `MYSQL_USER`/`TEST_MYSQL_USER`/etc env vars for their credentials. It doesn't make a difference to me, I just want to make sure we are aware of a pseudo-convention that is already in place.

Whichever approach is taken, I strongly agree that it needs to be non-interactive.

superkhau:
>The problem is bin/setup-mysql.js is written in such a way that it essentially requires someone to "press return" instead of a non-interactive mechanism.
>Whichever approach is taken, I strongly agree that it needs to be non-interactive.

I agree, if we're going to require the script be run, there should be an option for non-interactive mode.

bajtos:
> FYI, the connectors use a different convention and instead check for and use MYSQL_USER/TEST_MYSQL_USER/etc env vars for their credentials. It doesn't make a difference to me, I just want to make sure we are aware of a pseudo-convention that is already in place.

IIRC, I decided to not use those env vars because they are rather clumsy to configure on a local dev machine. It may also cause problems when there are multiple projects sharing the same database, let's say loopback-workspace and loopback-connector-mysql, especially when the connector tests require the database to be in a certain initial state IIRC.

I don't mind changing this decision, as long as we have an easy way how to setup local dev env.

However, I think my proposal in https://github.com/strongloop/loopback-workspace/pull/311#issuecomment-247989528 is the fastest way how to get CI green on Windows.

> I personally prefer modules to be a _little_ more turn-key and idempotent than that. Can we at least having `npm test` tell me (as a new contributor) what I need to do to get `npm test` to pass if I run it before running `bin/setup-mysql.js`?

The test suite is already printing an useful message when it cannot connect to MySQL, see https://github.com/strongloop/loopback-workspace/blob/3e80fcd3053199c5de549a5ded5e7773ad2fc5cc/test/end-to-end.js#L1059-L1085

```js
  connection.connect(function(err) {
    if (!err) return done(err);
    if (err.code === 'ECONNREFUSED') {
      err = new Error(
          'Cannot connect to local MySQL database, ' +
          'make sure you have `mysqld` running on your machine');
    } else {
      console.error();
      console.error('**************************************');
      console.error('Cannot connect to MySQL.');
      console.error('Setup the test environment by running');
      console.error('    node bin/setup-mysql');
      console.error('**************************************');
      console.error();
    }
    done(err);
  });
```

superkhau:
I vote for @bajtos suggestion at https://github.com/strongloop/loopback-workspace/pull/311#issuecomment-247989528 to get Windows green now.

rmg:
The reason for using env vars from CI instead of hard-coding `''` is so that no code has to be modified if CI needs to change credentials or any other aspect of DB connectivity.

Ideally every part of the connection string would be pulled from env vars, but from your proposal it wouldn't take much to extend it to use and env var with a fallback. eg:
```js
if (process.env.CI) {
  password = process.env.MYSQL_PASSWORD || '';
  return next();
}
```

rmg:
@slnode test please

test/end-to-end.js
```diff
@@ -29,8 +29,15 @@ var MYSQL_DATABASE = 'loopback_workspace_test';
 var MYSQL_USER = 'lbws';
 var MYSQL_PASSWORD = 'hbx42rec';
 
+before(function cleanTestDir(cb) {
+  fs.remove(PKG_CACHE, function(err) {
```
bajtos: Ouch, this completely cancels the point of using `strong-cached-install`. If we are going to delete `PKG_CACHE` before the tests run, then it's probably better to use `npm install` directly without `strong-cached-install`.

What problem is this trying to solve? Also how can be `PKG_CACHE` non-empty on CI machines, when each CI run starts in a clean cloned/checked-out repository?

I introduced `strong-cached-install` to speed up `npm test` when it's run repeatedly on a developer's machine. I.e. the first `npm test` is pretty slow, but subsequent `npm test` runs are much faster, as `npm install` is replaced with recursive copy.

Can we revert this change?

package.json
```diff
@@ -31,6 +31,7 @@
     "morgan": "^1.2.0",
     "ncp": "^2.0.0",
     "normalize-git-url": "^3.0.2",
+    "path": "^0.12.7",
```
raymondfeng: Why an external `path` dep?

bin/setup-mysql.js
```diff
@@ -17,23 +17,31 @@ USER = 'lbws';
 PASSWORD = 'hbx42rec';
 
 var connection;
+var password = '';
```
bajtos: nitpick: no need to initialise the variable here.

```js
var password, connection;
```

test/facet.js
```diff
@@ -8,6 +8,9 @@ var Facet = require('../').models.Facet;
 
 describe('Facet', function() {
   describe('Facet.create(def, cb)', function() {
+    this.timeout(15000);
+
+    before(resetWorkspace);
     beforeEach(givenBasicWorkspace);
```
bajtos: Isn't `givenBasicWorkspace` calling `resetWorkspace` under the hood? I find this change little bit suspicious, but if it fixes CI failures then I don't object.

test/model-definition.js
```diff
@@ -13,6 +13,7 @@ var ModelRelation = app.models.ModelRelation;
 var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var ConfigFile = app.models.ConfigFile;
+var nodePath = require('path');
```
bajtos: This is confusing, (almost all) Node.js code uses `var path = require('path')`. Can you please rename the other variables that are clashing with the `path` name?

test/model-definition.js
```diff
@@ -125,8 +126,9 @@ describe('ModelDefinition', function() {
       expect(path).to.equal(configFilePath);
     });
     it('should return construct configFile path', function() {
+      var modelPath = 'models/my-model.json';
       var path = ModelDefinition.getPath('.', { name: 'MyModel' });
```
bajtos: Use e.g. `modelPath` here so that `path` can be used for `require('path')`.

0candy:
Create new issue for the remaining windows failures for 404 errors.
https://github.com/strongloop/loopback-workspace/issues/328

PR #316, opened 2016-08-17
Remove Legacy Explorer

gunjpan:
- Removes `legacyExplorer` flag which
 enabled legacy end points: `/models` &
 `/routes`
- Update related tests & tests using the
legacyExplorer flag

Connect to https://github.com/strongloop/loopback/issues/2037

### DIFF
--- a/common/models/workspace.js
+++ b/common/models/workspace.js
@@ -215,6 +215,12 @@ module.exports = function(Workspace) {
             .filter(function(cc) { return cc.name != explorer; });
       }
 
+      // Add legacyExplorer flag to support creating LoopBack 2.x apps
+      if (loopbackVersion === '2.x' &&
+        template.server && template.server.config) {
+        template.server.config.push({ name: 'legacyExplorer', value: false });
+      }
+
       var dest = path.join(ConfigFile.getWorkspaceDir(), name);
       var steps = [];
 
--- a/server/server.js
+++ b/server/server.js
@@ -15,8 +15,6 @@ var boot = require('loopback-boot');
 var cookieParser = require('cookie-parser');
 var errorHandler = require('strong-error-handler');
 
-app.set('legacyExplorer', false);
-
 // required to support base models
 app.dataSource('db', {
   connector: loopback.Memory,
--- a/templates/projects/empty-server/data.js
+++ b/templates/projects/empty-server/data.js
@@ -82,7 +82,6 @@ template.server = {
       cors: false,
       handleErrors: false,
     }},
-    { name: 'legacyExplorer', value: false },
   ],
 
   modelConfigs: [
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -170,7 +170,6 @@ describe('end-to-end', function() {
             if (err) return done (err);
             var responseBody = JSON.stringify(res.body);
             expect(responseBody).to.include('stack');
-
             done();
           });
       });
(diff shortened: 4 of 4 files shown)


jannyHou:
The code change LGTM 👍

gunjpan:
@slnode test please

bajtos:
I am afraid this change is removing `legacyExplorer` setting for LoopBack 2.x applications too, which is not desired.

server/config.json
```diff
@@ -23,6 +23,5 @@
       "debug": true,
       "log": true
     }
-  },
-  "legacyExplorer": false
+  }
```
bajtos: loopback-workspace is still using LB 2.x under the hood, please revert.

server/server.js
```diff
@@ -15,8 +15,6 @@ var boot = require('loopback-boot');
 var cookieParser = require('cookie-parser');
 var errorHandler = require('strong-error-handler');
 
-app.set('legacyExplorer', false);
-
```
bajtos: This line is not needed because it's already set via `server/config.json`, it's ok to remove it.

templates/projects/empty-server/data.js
```diff
@@ -81,7 +81,6 @@ template.server = {
       cors: false,
       handleErrors: false,
     }},
-    { name: 'legacyExplorer', value: false },
```
bajtos: This should be parameterized. For apps based on LB 3.0 and newer, the entry should not be there. For apps based on LB 2.x, the entry should be preserved.

test/end-to-end.js
```diff
@@ -129,23 +129,6 @@ describe('end-to-end', function() {
...
-        .get('/api/models')
-        .expect(404, done);
-    });
-
-    it('has legacy explorer route /routes disabled', function(done) {
-      request(app)
-        .get('/api/routes')
-        .expect(404, done);
-    });
-
```
bajtos: I think it's better to keep these tests around. For 2.x apps, the routes should be disabled by config. For 3.x+ apps, the routes should not be there because they are not implemented at all. In all cases, these tests should pass.

bajtos:
@gunjpan left few comments to address, PTAL.

gunjpan:
@bajtos : Applied review feedback. PTAL, thanks.

common/models/workspace.js
```diff
@@ -215,6 +215,14 @@ module.exports = function(Workspace) {
             .filter(function(cc) { return cc.name != explorer; });
       }
 
+      // Remove legacyExplorer from LoopBack 3.x app template
```
bajtos: We should remove the option for versions beyond 3.x, e.g. 4.x too.

I think a better solution is to remove `legacyExplorer` from the template, and then add it here if `loopbackVersion === '2.x'`.

Thoughts?

bajtos:
@gunjpan great, the patch looks almost good now! I left one comment to address, also please check why all CI builds are failing.

common/models/workspace.js
gunjpan: Yup, I agree. It would be harder to manage for versions beyond 3.x. Will change it.

gunjpan:
@bajtos: Please review. Re. CI failures, it's mostly `timeout` issues, not sure how to deal with them.

test/end-to-end.js
```diff
@@ -525,6 +530,12 @@ describe('end-to-end', function() {
         .to.be.true;
       done();
     });
+
+    it('comes with legacyExplorer:false flag in config.json', function(done) {
+      var config = fs.readJsonSync(path.resolve(SANDBOX, 'server/config.json'));
+      expect(config.legacyExplorer).to.exist.and.to.be.false;
```
bajtos: Nitpick:

```js
expect(config).to.have.property('legacyExplorer', false)
```

https://github.com/moll/js-must#beware-of-libraries-that-assert-on-property-access

bajtos:
One more nitpick, the patch LGTM otherwise. No more review is needed, please squash the commits before merging.

gunjpan:
Merging -- as most CI failures are due to timeout, tests pass on local for sure.

PR #317, opened 2016-08-20
Added ElasticSearch

DOSarrest:
When trying to use loopback-connector-es with APIConnect Edit an error would occur when viewing data sources resulting in no datasources appearing at all. Upon troubleshooting the problem was found to be it was not in this loopback-workspace projects available-connectors.json file missing an entry to show es as an available source. The following pull request resolves this problem.

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -549,5 +549,45 @@
       }
     },
     "supportedByStrongLoop": true
+  },
+  {
+    "name": "es",
+    "description": "ElasticSearch",
+    "baseModel": "PersistedModel",
+    "features": {
+      "discovery": false,
+      "migration": false
+    },
+    "settings": {
+      "index": {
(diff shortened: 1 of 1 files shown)


slnode:
Can one of the admins verify this patch? To accept patch and trigger a build add comment ".*ok\W+to\W+test.*"

slnode:
Can one of the admins verify this patch?

slnode:
Can one of the admins verify this patch?

slnode:
Can one of the admins verify this patch?

slnode:
Can one of the admins verify this patch?

DOSarrest:
Is that commit linter error really a problem? The commit was auto generated from github.

bajtos:
Landed, thank you for the contribution!

PR #318, opened 2016-08-23
Fix empty-server template syntax.

richardpringle:
Fix the `empty-server` template to have consistent syntax.


richardpringle:
Currently `slc loopback` scaffolds a project with a particular `middleware.json` file. If you then do `slc loopback:model`, it will change the `middleware.json` file since its formatting is kind of weird in the first place. 

@0candy and @jannyHou, please take a look and tell me what you think.

jannyHou:
LGTM 🚢

Issue #319, opened 2016-08-30
slc loopback:model crashes when depending on git url


bajtos:
Reported by @azatoth in https://github.com/strongloop/strongloop/issues/316:
 
Having a dependency for "loopback" to "https://github.com/strongloop/loopback.git#2.x" results in `slc loopback:model` to crash with following error:

```
/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:678
    throw new TypeError('Invalid comparator: ' + comp);
    ^

TypeError: Invalid comparator: https://github.com/strongloop/loopback.git#2.x
    at Comparator.parse (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:678:11)
    at new Comparator (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:662:8)
    at /usr/lib/node_modules/strongloop/node_modules/semver/semver.js:780:12
    at Array.map (native)
    at Range.parseRange (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:779:13)
    at Range.<anonymous> (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:721:17)
    at Array.map (native)
    at new Range (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:720:40)
    at outside (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:1132:11)
    at Function.gtr (/usr/lib/node_modules/strongloop/node_modules/semver/semver.js:1126:10)

```

bajtos:
> Having a dependency for "loopback" to "https://github.com/strongloop/loopback.git#2.x" results in `slc loopback:model` to crash

@0candy PTAL. I think this is caused by this line: https://github.com/strongloop/loopback-workspace/blob/2cfab28e35b4ea7d962dc971339854eb7e678a4d/common/models/model-method.js#L17

```js
 return version != null ? !semver.gtr('3.0.0', version) : false;
```

0candy:
semver does not compare git url. Should set default loopback version if value is invalid for comparison.
https://github.com/strongloop/loopback-workspace/blob/master/server/connector.js#L218

PR #321, opened 2016-09-09
Fix semver error on git url

0candy:
Connect to https://github.com/strongloop/loopback-workspace/issues/319


0candy:
@bajtos  PTAL

lib/helper.js
```diff
@@ -0,0 +1,12 @@
...
+
+exports.DEFAULT_LB_VERSION = '2.x';
+
+/**
+ * validate loopback version
+ * @param version
+ * @returns valid loopback version or default if invalid
+ */
+exports.validateLoopbackVersion = function(version) {
+  return semver.validRange(version) == null ? this.DEFAULT_LB_VERSION : version;
```
bajtos: Using `this.DEFAULT_LB_VERSION` means the following code won't work:

```js
var validateLoopbackVersion = require('../lib/helper').validateLoopbackVersion;
validateLoopbackVersion(...);
```

I am proposing to use `exports.DEFAULT_LB_VERSION` or alternatively save the constant to a top-level variable.

```
var DEFAULT_LB_VERSION = '2.x';
exports.DEFAULT_LB_VERSION = DEFAULT_LB_VERSION;
//...
```

lib/helper.js
```diff
@@ -0,0 +1,12 @@
+var semver = require('semver');
+
+exports.DEFAULT_LB_VERSION = '2.x';
+
+/**
+ * validate loopback version
+ * @param version
+ * @returns valid loopback version or default if invalid
+ */
+exports.validateLoopbackVersion = function(version) {
```
bajtos: I think we spell LoopBack with capital `L` and `B`, see e.g. http://loopback.io/ or the caption of https://github.com/strongloop/loopback-workspace/blob/3e80fcd3053199c5de549a5ded5e7773ad2fc5cc/README.md#loopback-workspace-30.

Could you please rename the method to `validateLoopBackVersion`?

test/helper.js
```diff
@@ -0,0 +1,16 @@
...
+
+describe('helper', function() {
+  describe('validateLoopbackVersion()', function() {
+    it('should return valid loopback version', function() {
+      var version = helper.validateLoopbackVersion('2.1.3');
+      expect(version).to.equal('2.1.3');
+    });
+
+    it('should return default loopback version', function() {
+      var version = helper.validateLoopbackVersion('https://github.com/strongloop/loopback.git#2.x');
```
bajtos: Please add another test for `helper.validateLoopbackVersion('https://github.com/strongloop/loopback.git#2.x');`.

lib/helper.js
bajtos: Actually, since this is converting an npm dependency string into a `2.x`/`3.x` value, I think a name starting with `validate` is misleading. How about `parseLoopBackVersion` or `parseLoopBackDependency` or similar?

bajtos:
@0candy looks mostly good, I left few minor comments. The CI builds are failing on MacOS and Windows, I assume you checked the console outputs to verify the failures are not caused by this change.

0candy:
CI is failing because MySQL doesn't seem to be working again.

test/helper.js
bajtos: Sorry for the confusion, I meant 

```js
helper.validateLoopbackVersion('https://github.com/strongloop/loopback.git#1.x');
helper.validateLoopbackVersion('https://github.com/strongloop/loopback.git#master');
```

bajtos:
> CI is failing because MySQL doesn't seem to be working again.

That's annoying. Could you please chase down @rmg and find/learn what is the root cause of these MySQL failures? How much work is required to fix it?

rmg:
@bajtos @0candy the tests have never passed on Mac and Windows before. Recent changes to CI are just making this more visible.

rmg:
@slnode test please

0candy:
From Ryan:
The Mac and Windows environments were not being used for loopback-workspace until last week.
MySQL is not setup on those machines and Ryan will add as time permits.

rmg:
Correction, MySQL _is_ set up on these machines. The tests are failing on Windows because `package.json` uses a POSIX shell command (`test`) that is not supported by Windows.

lib/helper.js
```diff
@@ -0,0 +1,28 @@
...
+  // If its a valid loopback version, return it
+  var validRange = semver.validRange(version);
+  if (validRange != null) {
+    return version;
+  }
+  // Parse git url
+  var normalized = normalizeGitUrl(version);
+  var branch = normalized.branch;
+  // If its the master branch, return the master branch version.
+  if ((/master/i).test(branch)) {
```
bajtos: This will match any branch name containing `master`, case-insensitive. Is that intentional?

I would prefer to use a simpler condition `if (branch === 'master')`, what is the benefit of using a regular expression?

test/helper.js
```diff
@@ -0,0 +1,37 @@
+var helper = require('../lib/helper');
+var expect = require('chai').expect;
+
+describe('helper', function() {
+  describe('parseLoopBackVersion()', function() {
+    it('should return valid loopback version', function() {
```
bajtos: Please improve test names to only mention what's the outcome ("return valid loopback version"), but also what are the conditions when this outcome is expected. Most importantly, the test name should make it clear whether the input was a version, a version range, or a git url.

test/helper.js
```diff
@@ -0,0 +1,37 @@
...
+      var version = helper.parseLoopBackVersion('2.1.3');
```
bajtos: While testing a dependency version specified as exactly a single version may be useful, I think it's even more important to test version ranges: `^2.1.3`, `~2.1.3`, `2.x`, `2.1.x`.

bajtos:
> Correction, MySQL _is_ set up on these machines. The tests are failing on Windows because `package.json` uses a POSIX shell command (`test`) that is not supported by Windows.

I see, makes sense. @0candy I am ok to land this PR even with Windows builds failing, as long as you create a story to fix that ASAP, at latest in the next sprint.

Setting Windows builds aside, there are still few CI failures:
 - [cis-jenkins] nvm && dbs,0.12 — Failed! (45ff74f) [details](https://cis-jenkins.swg-devops.com/job/nb/job/loopback-workspace~master/label=nvm%20&&%20dbs,nodeVersion=0.12/158/)
 - MacOS builds - what's the problem there? `npm test` is regularly passing on our local MacOS machines.

0candy:
Created issue to fix CI for windows.
https://github.com/strongloop/loopback-workspace/issues/322

superkhau:
Suggested a possible solution at https://github.com/strongloop/loopback-workspace/issues/322#issuecomment-246792132

bajtos:
@0candy I think it's ok to land this patch despite the build failures. 🚢

0candy:
The nvm && dbs,0.10/0.12 are timeout errors which I'm taking care of in https://github.com/strongloop/loopback-workspace/pull/311

PR #326, opened 2016-09-27
Handle non-string values in parseLoopBackVersion

jtary:
Add handler to avoid a crash when parseLoopBackVersion is called with
no argument, or a non-string value.


jtary:
@bajtos PTAL

jtary:
Failures on windows VMs aren't related to this change. Merging.

PR #330, opened 2016-10-12
use casing specified in docs

eddiemonge:


bajtos:
@eddiemonge thank you for the pull request. While the documentation uses casing like `Boolean`, the runtime is case-insensitive and treats `boolean` the same way as `Boolean`.

In that light, I think this pull request is not adding any benefits, but instead it pollutes the git history.

Am I missing something here?

eddiemonge:
I was making it match the docs so anyone looking at this code as an example to make their own version of whatever else, don't get confused by the difference between docs and implementation. If you think thats not important then I am fine with closing this as it is just a small nice to have but not required

superkhau:
Can we update the docs to use lowercase instead of modify the code? I agree with @bajtos here as we should prefer the lowercase version anyways since it's supported.

@crandmck Can you give us a hand here?

crandmck:
> Can we update the docs to use lowercase instead of modify the code?

Boolean is always capitalized because it's named after George Boole (and so is a proper noun).  While I agree it would be nice to have consistency with code, in practice I don't think it will cause confusion.   Whether it's worth the cost to make the two consistent is up to you guys.  I guess there are a bunch of places where it's lower case (e.g. JSON in example repos).

superkhau:
>Boolean is always capitalized because it's named after George Boole (and so is a proper noun). While I agree it would be nice to have consistency with code, in practice I don't think it will cause confusion. Whether it's worth the cost to make the two consistent is up to you guys. I guess there are a bunch of places where it's lower case (e.g. JSON in example repos).

I think we should prefer common JavaScript conventions which is usually lowercasing for the string names. @crandmck We should use Uppercase when defining the datatypes directly (without quotes), as in `{type: Boolean}` vs `{type: 'boolean'}`.

crandmck:
To be clear, I was talking about usage in a sentence, where we follow standard English spelling rules (including capitalization).  In code examples (JS or JSON), we capitalize according to the code conventions.

Thus:

A Boolean is value is expressed as `{type: 'boolean'}`

So, I think we're in agreement.  I don't think the code casing needs to match the sentence spelling.

superkhau:
>A Boolean is value is expressed as {type: 'boolean'}

Yes, this is fine and I believe we're in agreement. I actually thought @eddiemonge was submitting a PR to lowercase the strings in quotes (rather than uppercasing).

@eddiemonge Do we need to update the  docs? I don't agree with the changes in this PR as @crandmck has the right conventions in the comments he posted above. I am proposing we reject this PR and get the docs to match. @crandmck @eddiemonge Can you paste a link to the doc itself?

eddiemonge:
It was like that in the old docs but the new ones seem mostly fixed except for this one: http://loopback.io/doc/en/lb2/LoopBack-types.html#array-types

crandmck:
Great, @eddiemonge pls open a PR in loopback.io for any changes to the above doc, thanks!

PR #332, opened 2016-10-13
Add key-value connectors

superkhau:
* kv-memory (builtin)

* loopback-connector-kv-redis

@bajtos PTAL

cc @0candy @jannyHou  

Connect to https://github.com/strongloop/loopback/issues/2771
Connect to https://github.com/strongloop/generator-loopback/issues/228

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -20,6 +20,16 @@
     "supportedByStrongLoop": true
   },
   {
+    "name": "kv-memory",
+    "description": "In-memory key-value connector",
+    "baseModel": "KeyValueModel",
+    "features": {
+      "discovery": false,
+      "migration": false
+    },
+    "supportedByStrongLoop": true
+  },
+  {
     "name": "db2",
     "description": "IBM DB2",
     "baseModel": "PersistedModel",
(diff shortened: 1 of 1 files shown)


available-connectors.json
```diff
@@ -191,6 +201,28 @@
...
+      "discovery": false,
+      "migration": false
+    },
+    "settings": {
+      "host": {
+        "type": "string"
+      },
+      "port": {
+        "type": "number"
+      }
```
bajtos: The Redis connector has more option than just host and port, see https://github.com/luin/ioredis#connect-to-redis and https://github.com/luin/ioredis/blob/master/API.md#new_Redis_new

 - `password` of type `string`
 - `db` of type `number`

We should also support `url`, but that's not implemented yet. I opened https://github.com/strongloop/loopback-connector-kv-redis/issues/15 for that.

superkhau:
@bajtos PTAL again

available-connectors.json
superkhau: Also added the `url` property because it seems like we'll need it for strongloop/loopback-connector-kv-redis#15.

superkhau: Also added `user` too

bajtos: @superkhau Could you please point me to a place where `user` and `pass` are documented? AFAICT from https://github.com/luin/ioredis/blob/master/API.md#new_Redis_new, the only authentication-related setting is `password`.

available-connectors.json
```diff
@@ -191,6 +201,42 @@
...
+      "host": {
+        "type": "string"
+      },
+      "port": {
+        "type": "number"
+      },
+      "user": {
+        "type": "string"
+      },
+      "pass": {
```
bajtos: Should be `password`, because that's the value expected by https://github.com/luin/ioredis/blob/master/API.md#new_Redis_new AFAICT.

superkhau:
cc @bajtos I need this landed for Monday to demo it. Gonna ask @0candy to take over review -- @0candy PTAL

superkhau:
@bajtos FYI, I landed with the "user" still there. I don't see it in the docs either (except in the url property), I can submit another PR to remove it if necessary.

available-connectors.json
```diff
@@ -191,6 +201,42 @@
...
+      },
+      "port": {
+        "type": "number"
+      },
+      "user": {
+        "type": "string"
+      },
+      "password": {
+        "type": "string",
+        "display": "password"
```
bajtos: Since the setting name is already `password`, there is no need for `display` to say the same again.

superkhau:
Follow up PR to fix the previous two review comments at https://github.com/strongloop/loopback-workspace/pull/336

PR #336, opened 2016-10-17
Remove unused properties for KV redis connector

superkhau:
- `user` property is not in the downstream connector docs

@bajtos PTAL

Connect to https://github.com/strongloop/loopback-workspace/pull/332

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -48,12 +48,8 @@
       "port": {
         "type": "number"
       },
-      "user": {
-        "type": "string"
-      },
       "password": {
-        "type": "string",
-        "display": "password"
+        "type": "string"
       },
       "database": {
         "type": "string"


available-connectors.json
```diff
@@ -48,12 +48,8 @@
       "port": {
         "type": "number"
       },
-      "user": {
-        "type": "string"
-      },
       "password": {
-        "type": "string",
-        "display": "password"
+        "type": "string"
```
bajtos: Oh, I think I was wrong here. I see that other connectors use `"display": "password"` too. Is this perhaps referring to Inquirer prompt types? 

https://www.npmjs.com/package/inquirer#password---type-password

screen shot 2016-10-18 at 14 52 03

@superkhau could you please check?

superkhau: It's required (if you don't put add the line, the user entry is not hidden with **** while you type). Reverting and landing.

superkhau:
Closing in favour of #338

Issue #331, opened 2016-10-13
discoverModelDefinition does not return on some tables


eddiemonge:
if `discoverModelDefinition` includes more than a certain amount of data (enough to cause a JSONified object to be larger than 8192b), then the process hangs.

its sending the data [here](https://github.com/strongloop/loopback-workspace/blob/68af105f184d86a5aeaa205e9b23db65d40939a6/bin/datasource-invoke.js#L87) but not being received [here](https://github.com/strongloop/loopback-workspace/blob/68af105f184d86a5aeaa205e9b23db65d40939a6/common/models/data-source-definition.js#L321).

Anything 8192b and above at that point and this hangs. The number of columns doesn't matter, its the amount of data. Not sure how to fix this since copying the conditions outside of loopback-workspace lets more data go through (up to something like 8mb).

bajtos:
@eddiemonge when copying links to code, please press `y` before copying the URL. When you press `y`, GitHub changes the URL to point to an exact commit, in other words a permalink is created that will work forever. The URLs you used in your comment are pointing to whatever code is currently in the `master` branch. Over time, the code changes and line numbers in the URL will eventually point to a different place.

I have edited the issue description and fixed the links.

sam-github:
@eddiemonge if you want me to look at this, you'll need to provide reproduction instructions. I have no time to read loopback-workspace, and figure out what it is, and how to call it.

eddiemonge:
@bajtos I normally do that. Not sure why I didn't this time. ¯\(°_o)/¯

sam-github:
https://github.ibm.com/eddie/loopback-workspace-test <-- from slack

sam-github:
@eddiemonge loopback-boot@2.23.1 doesn't exist, I'm trying to run with 2.23.0, but do I need to npm link in dev versions of some of the dependencies?

eddiemonge:
`2.23.0` should be fine. `2.23.1` is probably internal only at the moment but not sure why it would be

sam-github:
I get a response back for success and fail for node 4 and node 6 with loopback-workspace-test

sam-github:
I'm looking more closely at the code to see if I can comment on it without a repro.

eddiemonge:
PR posted with Sam's changes and some small tweaks

PR #337, opened 2016-10-18
Comment on forked child IPC

sam-github:
Comments on the code related to https://github.com/strongloop/loopback-workspace/issues/331

@eddiemonge

### DIFF
--- a/bin/datasource-invoke.js
+++ b/bin/datasource-invoke.js
@@ -22,9 +22,15 @@ process.once('message', function(msg) {
     send({
       error: err,
       callbackArgs: args,
-    });
-
-    process.nextTick(function() {
+    }, function() {
+      // XXX better to do this when the send has completed, which may be the
+      // actual source of your bug.
+      //
+      // But even better would be to not do this here. kill your child in your
+      // parent process after its received its message (assuming that you don't
+      // control the invoked methods... they shouldn't have side effects, and
+      // node should exit by itself, but if you don't trust them, best to kill
+      // from the parent where it cannot be stopped).
       process.exit();
     });
   }
@@ -68,12 +74,19 @@ function invoke(msg, cb) {
   }
 }
 
+// XXX(sam) don't do this, its unnecessary and dangerous, its not guaranteed
+// by node that this is valid, and the default behaviour of writing to stderr
+// is perfect for you, since you already collect the stderr in the parent.
 process.on('uncaughtException', function(err) {
+  // XXX(sam) the only way this is reachable is if globalize fails to
+  // initialize, AND this process doesn't have an IPC channel... delete all this
+  // code along with the uncaught exception handler.
   if (process.send) {
     send({
       error: err,
     });
   } else {
+    // XXX(sam) loses the stack trace :-(
     throw err;
   }
 });
@@ -86,6 +99,10 @@ function send(msg) {
   try {
     process.send(msg);
   } catch (e) {
+    // XXX(sam) get rid of this, too, the exception will already kill node,
+    // writing to stderr, and hit your top level stderr collector. There are way
+    // to many methods of sending errors to the parent in this code. Its un-unit
+    // testable, and unpredictable.
     g.error('failed to send message to parent process');
     console.error(e);
     process.exit(1);
--- a/common/models/data-source-definition.js
+++ b/common/models/data-source-definition.js
@@ -293,11 +293,16 @@ function ready(DataSourceDefinition) {
     http: { verb: 'POST' },
   });
 
+  // XXX(sam) This could really use a comment... it looks like the purpose here
+  // is not co-processing, but to workaround the fact that so much of what
+  // loopback does depends on the CWD from which its loaded?
   DataSourceDefinition.prototype.invokeMethodInWorkspace = function(methodName) {
     // TODO(bajtos) We should ensure there is never more than one instance
     // of this code running at any given time.
     var isDone = false;
     var self = this;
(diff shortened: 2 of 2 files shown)


common/models/data-source-definition.js
```diff
@@ -293,11 +293,16 @@ function ready(DataSourceDefinition) {
...
+  // XXX(sam) This could really use a comment... it looks like the purpose here
+  // is not co-processing, but to workaround the fact that so much of what
+  // loopback does depends on the CWD from which its loaded?
   DataSourceDefinition.prototype.invokeMethodInWorkspace = function(methodName) {
     // TODO(bajtos) We should ensure there is never more than one instance
     // of this code running at any given time.
     var isDone = false;
     var self = this;
+    // XXX(sam) why not call slice as (arguments, 1) instead of shifting off
+    // arguments[0] later?
```
eddiemonge: no idea but that is a good point

bin/datasource-invoke.js
```diff
@@ -22,9 +22,15 @@ process.once('message', function(msg) {
...
-    process.nextTick(function() {
+    }, function() {
+      // XXX better to do this when the send has completed, which may be the
+      // actual source of your bug.
+      //
+      // But even better would be to not do this here. kill your child in your
+      // parent process after its received its message (assuming that you don't
+      // control the invoked methods... they shouldn't have side effects, and
+      // node should exit by itself, but if you don't trust them, best to kill
+      // from the parent where it cannot be stopped).
```
bajtos: Many LoopBack connectors create a connection pool at setup time, I think that's why we added `process.exit()` call here.

> kill your child in your parent process after its received its message

Sounds like the right approach to me.

bin/datasource-invoke.js
```diff
@@ -68,12 +74,19 @@ function invoke(msg, cb) {
   }
 }
 
+// XXX(sam) don't do this, its unnecessary and dangerous, its not guaranteed
+// by node that this is valid, and the default behaviour of writing to stderr
+// is perfect for you, since you already collect the stderr in the parent.
 process.on('uncaughtException', function(err) {
+  // XXX(sam) the only way this is reachable is if globalize fails to
+  // initialize, AND this process doesn't have an IPC channel... delete all this
+  // code along with the uncaught exception handler.
```
bajtos: IIRC, the idea here is that if the LoopBack app cannot be loaded because the connection configuration of a datasource is invalid, we forward the error message to the caller, so that the caller can provide a specific error code for the UI, so that UI can show a useful error message like "cannot connect to the database" instead of a generic "datasource-invoke crashed".

Having said that, I think parsing the stdout/stderr output to extract the error message from there is a viable strategy too. Especially if the current implementation is not reliable enough.

common/models/data-source-definition.js
```diff
@@ -293,11 +293,16 @@ function ready(DataSourceDefinition) {
     http: { verb: 'POST' },
   });
 
+  // XXX(sam) This could really use a comment... it looks like the purpose here
+  // is not co-processing, but to workaround the fact that so much of what
+  // loopback does depends on the CWD from which its loaded?
```
bajtos: Workspace is a LoopBack app itself. We need to run the inspected LoopBack app out-of-process for multiple reasons:
 - prevent interference between workspace and app models
 - prevent `require.resolve` caching when we want load a new revision of the app, after we made changes
 - prevent the workspace app from crashing if the LoopBack app crashes (e.g. because it cannot load a newly added connector that was not `npm install`-ed yet)

common/models/data-source-definition.js
bajtos: I personally don't mind too much.

Note that we are doing more than just `args.shift()`, there is also `args.pop()`. I think the idea here is that we start with `args` as a write-able Array copy of our arguments that we modify later as needed.

common/models/data-source-definition.js
```diff
@@ -310,6 +315,9 @@ function ready(DataSourceDefinition) {
       cb = args.pop();
     } else {
       cb = function invokeComplete(err) {
+        // XXX(sam) when someone sees this, they won't know where its logged
+        // from, maybe use console.trace(), or put some kind of greppable string
+        // as first arg to console.error
```
bajtos: Makes sense 👍

common/models/data-source-definition.js
```diff
@@ -327,9 +335,19 @@ function ready(DataSourceDefinition) {
       done.apply(self, msg.callbackArgs);
     });
 
+    // XXX(sam) use https://www.npmjs.com/package/concat-stream
     child.stderr.on('data', storeErrors);
 
-    child.on('exit', function(code) {
+    // XXX(sam) don't use exit here, its racy. The child can exit while its data
+    // is still unread from the pipes, in which case stderr will not be
+    // complete. Use 'close', it exists for this purpose, and has the same args.
```
bajtos: Makes sense, :+1:

bin/datasource-invoke.js
sam-github: Agreed, and there is already parsing code, its just that its parsing text it gets from json IPC message, rather than text it gets from stderr.

bin/datasource-invoke.js
sam-github: That sortof makes sense... except that connection pools should be unrefed, I think, so they don't hold node open if there is no server listening. But that is probably a fantasy world, there will always be bugs, its not a bad idea to kill from parent.

eddiemonge:
landed changes in pull request

PR #340, opened 2016-10-19
change when child process exits

### LINKED ISSUE #331: discoverModelDefinition does not return on some tables

eddiemonge:
Process should exit when send is complete. Also removed extra code exits
as they aren't necessary.

Took the comments from Sam at https://github.com/strongloop/loopback-workspace/pull/337

Fixes #331

### DIFF
--- a/bin/datasource-invoke.js
+++ b/bin/datasource-invoke.js
@@ -12,23 +12,13 @@ var assert = require('assert');
 process.once('message', function(msg) {
   invoke(msg, function(err) {
     if (err) {
-      if (!err.origin) err.origin = 'invoke';
-      return done(err);
+      err.origin = err.origin || 'invoke';
     }
-    done(null, Array.prototype.slice.call(arguments, 0));
+    done(err, Array.prototype.slice.call(arguments, 0));
   });
-
-  function done(err, args) {
-    send({
-      error: err,
-      callbackArgs: args,
-    });
-
-    process.nextTick(function() {
-      process.exit();
-    });
-  }
 });
+
+process.on('uncaughtException', done);
 
 function invoke(msg, cb) {
   var dataSourceName = msg.dataSourceName;
@@ -68,31 +58,22 @@ function invoke(msg, cb) {
   }
 }
 
-process.on('uncaughtException', function(err) {
-  if (process.send) {
-    send({
-      error: err,
-    });
-  } else {
+function done(err, args) {
+  if (!process.send) {
     throw err;
   }
-});
 
-function send(msg) {
-  if (msg.error) {
-    msg.error = toSerializableError(msg.error);
-  }
-
-  try {
-    process.send(msg);
-  } catch (e) {
-    g.error('failed to send message to parent process');
-    console.error(e);
-    process.exit(1);
-  }
+  process.send({
+    error: toSerializableError(err),
+    callbackArgs: args,
+  });
 }
 
 function toSerializableError(err) {
+  if (!err) {
+    return null;
+  }
+
   var alt = {};
 
   Object.getOwnPropertyNames(err).forEach(function(key) {
--- a/common/models/data-source-definition.js
+++ b/common/models/data-source-definition.js
@@ -325,6 +325,7 @@ function ready(DataSourceDefinition) {
       }
 
       done.apply(self, msg.callbackArgs);
+      child.kill();
     });
 
     child.stderr.on('data', storeErrors);


sam-github:
Did you verify this fixes #331? Large json gets sent across OK now?

eddiemonge:
Yes. Tested small and big

eddiemonge:
Not sure why CI is failing

eddiemonge:
@slnode test please

eddiemonge:
@slnode test please

jtary:
LGTM

superkhau:
I don't see any issues, but I'll leave it to @0candy @jannyHou for final LGTM as they are the FA owners for this repo.

sam-github:
LGTM, when you said it works, I figured  it was a done deal

0candy:
@eddiemonge The error on cis-jenkins is valid.
```child_process.js:431
        throw new TypeError("This handle type can't be sent");
              ^
TypeError: This handle type can't be sent
    at process.target.send (child_process.js:431:15)
    at process.done (/home/jenkins/workspace/nb/loopback-workspace~master/087d56b6/bin/datasource-invoke.js:67:11)
    at process.emit (events.js:95:17)
    at process._fatalException (node.js:310:26)```

On node 10 & 12 `process.send()` was synchronous and doesn't accept a callback function.

**Comment by eddiemonge** [comment]
Does that mean it needs a conditional check to see which version of node it is running on?

**Comment by sam-github** [comment]
Yes. At which point you may be lucky and on 0.10 and 0.12 the send is sync, so you never lose data.

Or you may get unlucky and need to implement another mechanism, like kill of child from parent after the entire message has been received.

File: bin/datasource-invoke.js [hunk 1]
```diff
@@ -17,19 +17,10 @@ process.once('message', function(msg) {
...
-      callbackArgs: args,
-    });
-
-    process.nextTick(function() {
-      process.exit();
-    });
-  }
 });
 
+process.on('uncaughtException', done);
```
eddiemonge: testing

superkhau: change please

superkhau:
Ignore my comments, above was playing with review comments with @eddiemonge.

bin/datasource-invoke.js
```diff
@@ -8,28 +8,19 @@
...
 var g = require('strong-globalize')();
 var assert = require('assert');
+var semver = require('semver');
 
 process.once('message', function(msg) {
   invoke(msg, function(err) {
     if (err) {
-      if (!err.origin) err.origin = 'invoke';
-      return done(err);
+      err.origin = err.origin || 'invoke';
```
superkhau: return done(err);

bin/datasource-invoke.js
```diff
@@ -8,28 +8,19 @@
...
     }
-    done(null, Array.prototype.slice.call(arguments, 0));
+    done(err, Array.prototype.slice.call(arguments, 0));
```
superkhau: You can leave this line alone with the above change.

superkhau: Nevermind, I see what's going on below. Ignore comments above.

common/models/data-source-definition.js
```diff
@@ -325,6 +325,7 @@ function ready(DataSourceDefinition) {
       }
 
       done.apply(self, msg.callbackArgs);
+      child.kill();
```
superkhau: Is this necessary?

eddiemonge: yeah since the child isn't smart enough to figure it out when it should exit. which is the reason for this whole pr.

rmg:
@slnode test please

superkhau:
@eddiemonge You can be able to downstream ignore apim-ui (if you deem it necessary, you are the SME here). See https://github.com/strongloop/loopback-datasource-juggler/blob/master/package.json#L56-L62 -- I don't think we can target direct versions, only full repo names in the list ATM, you can try it out and confirm if you want).

bin/datasource-invoke.js
```diff
@@ -8,28 +8,19 @@
 
 var g = require('strong-globalize')();
 var assert = require('assert');
+var semver = require('semver');
```
bajtos: I think this is no longer used and should be removed.

superkhau:
Overriding merge. @eddiemonge Can you submit a PR to downstream ignore apim-ui if you don't need it to pass for workspace PRs?

eddiemonge:
we need apim-ui but not the old versions. @rmg is there a way to ignore specific builds of it?

Issue #328, opened 2016-09-29
Test case failures on windows


0candy:
14 test cases are throwing 404 instead of the expected status code on windows.

```
1) end-to-end empty-server template omits sensitive error details in production mode:
[...]

PR #342, opened 2016-11-09
Validate upserts

bajtos:
Enable `validateUpsert` on all models (workspace entities) to ensure validation is triggered for calls made by front-end UI (APIC Editor), which use `PUT` almost exclusively.

This should fix the problem where APIC Editor allows users to create a property name containing a dot, even though such value is not a valid name according to ModelProperty validation rules.

@jannyHou please review

@anthonyettinger or @eddiemonge could you please `npm install strongloop/loopback-workspace#fix/upsert-validation` in your APIC Editor codebase and do a quick check to verify that this change is not breaking any existing functionality?

### DIFF
--- a/test/model-property.js
+++ b/test/model-property.js
@@ -9,10 +9,12 @@ var ModelProperty = app.models.ModelProperty;
 var ModelDefinition = app.models.ModelDefinition;
 var ConfigFile = app.models.ConfigFile;
 var TestDataBuilder = require('./helpers/test-data-builder');
+var request = require('supertest');
 
 describe('ModelProperty', function() {
   beforeEach(givenBasicWorkspace);
   beforeEach(function(done) {
+    this.modelId = 'server.user';
     ModelDefinition.create({
       name: 'user',
       facetName: 'server',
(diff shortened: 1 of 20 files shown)


test/model-property.js
```diff
@@ -9,10 +9,12 @@ var ModelProperty = app.models.ModelProperty;
 var ModelDefinition = app.models.ModelDefinition;
 var ConfigFile = app.models.ConfigFile;
 var TestDataBuilder = require('./helpers/test-data-builder');
+var request = require('supertest');
 
 describe('ModelProperty', function() {
   beforeEach(givenBasicWorkspace);
   beforeEach(function(done) {
+    this.modelId = 'server.user';
```
jannyHou: any reason we use `this.modelId` here instead of a `var modelId = 'server.user'`? also noticed other similar format in this test file.

jannyHou:
Only one question, the pr LGTM 🚢 

[cis-jenkins] downstream apim-ui failures: Due to npm install failure, not related to this pr.
loopback-workspace windows failure: I believe it's the same as https://github.com/strongloop/loopback-workspace/issues/328#issuecomment-258987924

test/model-property.js
bajtos: Good point. `this.modelId` will store the data on a "context" object passed as `this` to all Mocha test/hook callbacks. We could use function scopes and `var modelId` instead, that would work equally well here.

I used `this.modelId` for consistency with the existing code around.

Issue #348, opened 2016-11-21
read artifacts from a workspace directory


deepakrkris:
Feature: Users should be able to read a list of artifacts to reference the contents of my application by name
- [x] Get a list of models
- [x] find a model definition
- [x] find a model property
- [x] find a model method
- [ ] find a model relation
- [ ] get a list of access control list (ACL) for a model
- [ ] get a list of middleware phases
- [ ] find a middleware method

PR #351, opened 2016-11-22
add ability to read model property config

deepakrkris:
Iteration 1 : read a model property config from model definition

#348

### DIFF
--- a/common/models/model-property.js
+++ b/common/models/model-property.js
@@ -1,0 +1,59 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-workspace
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+'use strict';
+var fs = require('fs-extra');
+var path = require('path');
+
+/**
+  * Represents a Property of a LoopBack `Model`.
+  *
+  * @class ModelProperty
+  * @inherits WorkspaceEntity
+  */
+module.exports = function(ModelProperty) {
+  ModelProperty.validatesFormatOf('name', {with: /^[\-_a-zA-Z0-9]+$/});
+
+  /**
+   * List of built-in types that can be used for `ModelProperty.type`.
+   * @type {string[]}
+   */
+  ModelProperty.availableTypes = [
+    'string',
+    'number',
+    'boolean',
+    'object',
+    'array',
+    'date',
+    'buffer',
+    'geopoint',
+    'any',
+  ];
+
+  ModelProperty.getAvailableTypes = function(cb) {
+    cb(null, ModelProperty.availableTypes);
+  };
+
+  ModelProperty.remoteMethod('getAvailableTypes', {
+    http: {verb: 'get', path: '/available-types'},
+    returns: {type: ['string'], root: true},
+  });
+
+  ModelProperty.find = function(workspaceDir, id, cb) {
+    var parts = id.split('.');
+    var facet = parts[0];
+    var modelName = parts[1];
+    var propertyName = parts[2];
+    var file = path.resolve(workspaceDir, modelName + '.json');
+    fs.readJson(file, function(err, data) {
+      if (err && err.name === 'SyntaxError') {
+        err.message = g.f('Cannot parse %s: %s', id, err.message);
+        return cb(err);
+      }
+      var properties = data.properties;
+      var propertyConfig = properties[propertyName];
+      cb(null, propertyConfig);
+    });
+  };
+};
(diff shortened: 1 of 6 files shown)


common/models/model-property.js
```diff
@@ -0,0 +1,60 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-workspace
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+'use strict';
+
+var path = require('path');
+var fs = require('fs-extra');
```
superkhau: path goes under fs -- lexical ordering prefered

common/models/model-property.js
```diff
@@ -0,0 +1,60 @@
...
+
+module.exports = function(ModelProperty) {
+  /**
+   * Represents a Property of a LoopBack `Model`.
+   *
+   * @class ModelProperty
+   * @inherits WorkspaceEntity
+   */
```
superkhau: This comment should go on line 9 right above module exports

superkhau: We should also name struture the file differently -- i'm thinking like:

```
module.exports = ModelProperty;


function ModelProperty(..) {
}

ModelProperty.getAvailableTypes = function(...) {
}

...

and so forth to prevent the nesting.

File: test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js [hunk 3]
```diff
@@ -0,0 +1,39 @@
+'use strict';
+var util = require('util');
+var async = require('async');
+var app = require('../../../../../');
+var loopback = require('loopback');
+var chai = require('chai');
```
superkhau: var expect = require('chai').expect;

Ideally it should be moved out to `test/helpers/expect.js` and simply `var expect = require('../helpers/expect');` here.

test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js
```diff
@@ -0,0 +1,39 @@
...
+var path = require('path');
+var ModelProperty = app.models.ModelProperty;
+var exampleWorkspace = path.resolve(__dirname, '../../../../../example/common/models');
+
+module.exports = function() {
+  var testcase = this;
```
superkhau: get rid of testcase

```
var propertyName;
var modelName;

this.Given..., function(model, prop, next) {
  modelName = model;
  propertyName = prop;
  next();
});

File: test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js [hunk 5]
```diff
@@ -0,0 +1,39 @@
+'use strict';
+var util = require('util');
+var async = require('async');
+var app = require('../../../../../');
+var loopback = require('loopback');
+var chai = require('chai');
+var path = require('path');
+var ModelProperty = app.models.ModelProperty;
+var exampleWorkspace = path.resolve(__dirname, '../../../../../example/common/models');
```
superkhau: Is there a before() hook you can do these in so we don't lump them together with the imports?

test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js
```diff
@@ -0,0 +1,39 @@
...
+
+module.exports = function() {
+  var testcase = this;
+  this.Given(/^The model '(.+)' has a property '(.+)'$/, function(modelName, propertyName, next) {
+    testcase.modelName = modelName;
+    testcase.propertyName = propertyName;
```
superkhau: Something is weird here, you get the modelNames and properties but don't use it down below.

common/models/model-property.js
deepakrkris: agreed

common/models/model-property.js
deepakrkris: agreed

test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js
deepakrkris: helpers would be added in the next iteration. this iteration is only to add basic test cases

test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js
deepakrkris: there is no before() hook in cucumber

test/acceptance-tests/workspace-devtime/step-definitions/reading-artifacts/findModelProperty.js
deepakrkris: agreed

common/models/model-property.js
deepakrkris: I see that there is no way to avoid this nesting in loopback !

superkhau:
Use a real datagraph instead of readJson.

PR #358, opened 2016-12-15
return after err done in tests

eddiemonge:
If there was an error in the test, `done` was called and then _the rest of the code in the test still ran_. This includes any additional `done`s. This led to errors being thrown about multiple `done`s being called.

### DIFF
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -55,7 +55,7 @@ describe('end-to-end', function() {
       request(app)
         .get('/')
         .expect(200, function(err, res) {
-          if (err) done(err);
+          if (err) return done(err);
           expect(res.body).to.have.property('uptime');
           done();
         });
@@ -264,7 +264,7 @@ describe('end-to-end', function() {
       request(app)
         .get('/')
         .expect(200, function(err, res) {
-          if (err) done(err);
+          if (err) return done(err);
           expect(res.body).to.have.property('uptime');
           done();
         });
@@ -408,13 +408,9 @@ describe('end-to-end', function() {
       .post('/api/Notes')
       .send(sample)
       .expect(200, function(err, res) {
-        if (err) {
-          done(err);
-        } else {
-          expect(res.body).
-            to.have.property('title', 'myTitle');
-          done();
-        };
+        if (err) return done(err);
+        expect(res.body).to.have.property('title', 'myTitle');
+        done();
       });
     });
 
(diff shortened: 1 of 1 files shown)


test/end-to-end.js
```diff
@@ -408,13 +408,10 @@ describe('end-to-end', function() {
...
-        if (err) {
-          done(err);
-        } else {
-          expect(res.body).
-            to.have.property('title', 'myTitle');
-          done();
-        };
+        if (err) return done(err);
+        expect(res.body).
+          to.have.property('title', 'myTitle');
```
superkhau: Does it fit on one line wrapped at 80 chars?

test/end-to-end.js
```diff
@@ -424,13 +421,10 @@ describe('end-to-end', function() {
...
```
superkhau: ditto

test/end-to-end.js
```diff
@@ -443,13 +437,10 @@ describe('end-to-end', function() {
...
-            if (err) {
-              done(err);
-            } else {
-              expect(res.body).
-                to.have.property('count', 1);
-              done();
-            }
+            if (err) return done(err);
+            expect(res.body).
+              to.have.property('count', 1);
```
superkhau: ditto

eddiemonge:
Even though this makes the tests pass, I think its the wrong solution. There shouldn't be errors to begin with unless it is testing the error case, which I don't think it is. I think this might be better in these cases to `expect(err).to.be.undefined`

test/end-to-end.js
superkhau: Why not, leave the house cleaner. If you're really nitpicky, separate commit should be good enough.

superkhau:
We use `if (err) return done(err)` for better stack traces as per @bajtos. I agree with you though, `expect(err).to.be.undefined()` -- with dirty-chai -- makes more sense to me (one gotcha is sometimes err comes back as `null` though for some libs, which I find annoying -- return done(err) doesn't run into those issues).

eddiemonge:
`expect(!!err).to.be.false` then. `return done(err)` doesn't seem to mark the test as failed.

superkhau:
`expect(!!err).to.be.false` seems hackish, return done(err) should return standard err -- throwing during mocha tests. Its not working because there is a `process.on('unhandledException')` in the code there.

eddiemonge:
@superkhau so is this good as is then?

superkhau:
@eddiemonge LGTM. :ship:

eddiemonge:
guess i can't merge this. no access. @deepakrkris @0candy @jannyHou

jannyHou:
no permission either :(

superkhau:
@eddiemonge Only admin's/squad leads should be doing overrides -- https://github.com/strongloop-internal/loopback-knowledge-base/blob/master/ground-rules.md#overrides.

That said, I can override for you.

superkhau:
We discussed with @ritch during standup that this should be safe to land since we're not releasing. @bajtos Can you take a look later and confirm if we need to send a fixup PR for any comments you may have? I'm override merge now.

bajtos:
This patch is great, thank you for making these fixes! 👏

PR #359, opened 2016-12-15
Add package.json

deepakrkris:
Add initial package description and dev dependencies

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,2 @@
+var loopback = require('loopback');
+var app = module.exports = loopback();
--- a/package.json
+++ b/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "loopback-workspace",
+  "version": "4.0.0-alpha.1",
+  "main": "index.js",
+  "scripts": {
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "loopback-workspace",
+  "version": "4.0",
```
rmg: Version must be valid semver.

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "scripts": {
+    "lint": "eslint ."
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/strongloop/loopback-workspace.git"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "loopback": "latest"
```
rmg: This should not be `latest`.

index.js
```diff
@@ -0,0 +1,3 @@
+var loopback = require('loopback');
+
+var app = module.exports = loopback();
```
rmg: Missing terminating EOL.

PR #365, opened 2016-12-22
4.x : Workspace Refactoring : Feature to add a datasource definition

deepakrkris:
Create a datasource node in the workspace graph

- [X] - define a DataSource Class
- [X] - add a DataSourceDefinition loopback model
- [X] - add workspace task to add a Datasource node to a workspace graph
- [X] - extend connector to create a datasource
- [X] - add gherkin cucumber test case to create a datasource
- [X] - use Dirty Chai
- [X] - move graph modules in datamodel to sub directory 'graph'

### DIFF
--- a/common/models/data-source-definition.js
+++ b/common/models/data-source-definition.js
@@ -1,0 +1,25 @@
+// Copyright IBM Corp. 2015,2016. All Rights Reserved.
+// Node module: loopback-workspace
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+'use strict';
+
+module.exports = function(DataSourceDefinition) {
+  /**
+   * Creates a data source definition.
+   *
+   * @class DataSourceDefinition
+   */
+  DataSourceDefinition.on('dataSourceAttached', function() {
+    DataSourceDefinition.create = function(data, options, cb) {
+      if (typeof options === 'function') {
+        cb = options;
+        options = null;
+      }
+      var connector = DataSourceDefinition.getConnector();
+      var id = data.id;
+      //TODO(Deepak) - add response handling later as part of the callback
+      connector.createDataSource(id, data, cb);
+    };
+  });
+};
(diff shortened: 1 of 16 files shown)


common/models/data-source-definition.js
```diff
@@ -0,0 +1,25 @@
...
+'use strict';
+
+module.exports = function(DataSourceDefinition) {
+  /**
+   * Creates a DataSource definition.
+   *
+   * @class DataSourceDefinition
+   */
+  DataSourceDefinition.on('dataSourceAttached', function(eventData) {
+    var connector = DataSourceDefinition.getConnector();
```
superkhau: move this down to line 21 -- where it's used

component/tasks.js
```diff
@@ -13,6 +14,11 @@ class Tasks {
     var model = new Model(workspace, modelId, modelDef);
     cb(null, modelDef);
   }
+  addDataSource(id, datasource, cb) {
+    var workspace = this;
+    var ds = new DataSource(workspace, id, datasource);
+    cb(null, datasource);
```
superkhau: we don't want to pass back ds?

test/acceptance-tests/workspace-devtime/step-definitions/create-datasource.js
```diff
@@ -0,0 +1,42 @@
...
+      };
+      DataSourceDefinition.create(datasource, {}, function(err, data) {
+        if (err) return next(err);
+        testsuite.expectedDs = datasource;
+        next();
+      });
+    });
+
+  this.Then(/^the datasource definition is created$/, function(next) {
+    var expect = chai.expect;
```
superkhau: Move out to shared helper -- see https://github.com/strongloop/loopback-connector-kv-redis/blob/master/test/helpers/expect.js

superkhau:
In addition to all the above? Are you intending to use ES6? ie. const + let instead of var?

common/models/data-source-definition.js
deepakrkris: comment

component/tasks.js
deepakrkris: we cannot instantiate without a variable

common/models/data-source-definition.js
deepakrkris: test

PR #372, opened 2017-01-03
Fix handleErrors config in the template

bajtos:
Fix "empty-server" template to correctly disable the built-in error handler provided by REST adapter in strong-remoting.

See https://github.com/strongloop/loopback.io/pull/134 and https://github.com/strongloop/loopback.io/pull/134#issuecomment-270097859 in particular for background information.

@superkhau please review
cc @doublemarked

### DIFF
--- a/templates/projects/empty-server/data.js
+++ b/templates/projects/empty-server/data.js
@@ -67,6 +67,7 @@ template.server = {
     { name: 'remoting', value: {
       context: false,
       rest: {
+        handleErrors: false,
         normalizeHttpPath: false,
         xml: false,
       },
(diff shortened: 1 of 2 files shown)


bajtos:
Also cc @loay since this bug was introduced by his pull request #292 (just FYI).

superkhau:
@slnode test please

superkhau:
@anthonyettinger @eddiemonge Is apim-ui always failing? It is a good candidate for downstream ignoring?

bajtos:
Released as `loopback-workspace@3.36.1`.

PR #379, opened 2017-01-09
4.x : Workspace Refactoring : Add loopback model to read middleware

deepakrkris:
Functionality to read middleware configuration from workspace


common/models/middleware.js
```diff
@@ -16,6 +16,23 @@ module.exports = function(Middleware) {
     }
     return phase;
   };
+  Middleware.getPhaseFromId = function(id) {
+    const parts = id.split('.');
+    if (parts.length && parts.length > 1) {
```
eddiemonge: parts will always have length here since its an array. If it doesn't it will have thrown an error in the split

common/models/middleware.js
```diff
@@ -16,6 +16,23 @@ module.exports = function(Middleware) {
...
+      let phase = parts[0];
```
eddiemonge: 22-26 could be rewritten as:

`return parts[0] + (parts.length > 2 ? ':' + parts[1] : '');`

common/models/middleware.js
```diff
@@ -16,6 +16,23 @@ module.exports = function(Middleware) {
...
+      if (parts.length > 2) {
+        phase = phase + ':' + parts[1];
+      }
+      return phase;
+    }
```
eddiemonge: should this return undefined if there isnt a period in the string?

common/models/middleware.js
```diff
@@ -16,6 +16,23 @@ module.exports = function(Middleware) {
...
+  };
+  Middleware.getMiddlewarePath = function(id) {
+    const parts = id.split('.');
+    if (parts.length && parts.length > 1) {
```
eddiemonge: `return parts.length > 1 ? parts.pop() : id;`

component/datamodel/middleware-phase.js
```diff
@@ -28,6 +28,17 @@ class MiddlewarePhase extends Node {
     }
     return {};
   }
+  getMiddleware(middlewarePath) {
+    const contains = this._contains;
```
eddiemonge: this function could be rewritten as 

```
  const middlewares = this._contains.Middleware;
  const pointer = middlewares && middlewares[middlewarePath];
  return pointer && pointer.getNode();
```

component/datamodel/util/file-utility.js
```diff
@@ -111,3 +112,12 @@ function readDataSource(workspace, cb) {
     cb(null, data);
   });
 }
+
+function readMiddleware(workspace, cb) {
+  const filePath = workspace.getMiddlewareFilePath();
+  fs.readJson(filePath, function(err, data) {
+    if (err) return err;
```
eddiemonge: this seems like an odd pattern. That means usage would look like:

```
var err = readMiddleWare('ws', function(iDontKnow, data) {
  // stuff that will never use the first param of the callback function
});
if (err) (console.error('there was an error outside of the callback');
```

Seems like it should call the callback instead of returning err. Kind of weird that the return of the function is true if there is an error and undefined if actually passes.

component/tasks.js
```diff
@@ -89,6 +89,10 @@ class Tasks {
     const workspace = this;
     fsUtility.readDataSource(workspace, cb);
   }
+  refreshMiddleware(cb) {
+    const workspace = this;
```
eddiemonge: storing `this` looks unnecessary since its not used in a callback or more than once

component/workspace.js
```diff
@@ -123,6 +123,25 @@ class Workspace extends Graph {
     }
     return config;
   }
+  setMiddlewareConfig(config) {
+    const workspace = this;
+    const phases = this.middlewarePhases;
```
eddiemonge: unused variable

component/workspace.js
```diff
@@ -123,6 +123,25 @@ class Workspace extends Graph {
     }
     return config;
   }
+  setMiddlewareConfig(config) {
+    const workspace = this;
```
eddiemonge: if you use arrow functions you dont need this variable

connector/index.js
```diff
@@ -102,3 +102,8 @@ connector.findDataSource = function(id, cb) {
   const workspace = WorkspaceManager.getWorkspace();
   DataSourceHandler.findDataSource(workspace, id, cb);
 };
+
+connector.findMiddleware = function(phase, name, cb) {
+  const workspace = WorkspaceManager.getWorkspace();
+  MiddlewareHandler.findMiddleware(workspace, phase, name, cb);
```
eddiemonge: should this be returned? along with line 30 in middleware-handler?

connector/middleware-handler.js
```diff
@@ -13,5 +13,21 @@ class MiddlewareHandler {
     const taskList = [create];
     workspace.execute(taskList, callback);
   }
+  static findMiddleware(workspace, phaseName, middlewarePath, cb) {
+    function refresh(next) {
+      workspace.refreshMiddleware(function(err) {
```
eddiemonge: would `workspace.refreshMiddleware(next);` not work here?

connector/middleware-handler.js
```diff
@@ -13,5 +13,21 @@ class MiddlewareHandler {
...
+        next(err);
+      });
+    }
+    function callback(err, results) {
+      if (err) return cb(err);
+      const phase = workspace.getMiddlewarePhase(phaseName);
+      const middleware = phase.getMiddleware(middlewarePath);
+      if (middleware) return cb(null, middleware.getConfig());
+      cb('middleware not found');
```
eddiemonge: i think this is supposed to go through strong globalize maybe?

common/models/middleware.js
deepakrkris: yes , it should return undefined

connector/middleware-handler.js
deepakrkris: globalization would be added later

common/models/middleware.js
deepakrkris: when run independent, the following code does not throw an error
```const id = 'models';
const parts = id.split('.');
if (parts.length && parts.length > 1) {
  console.log(parts[1]);
} else {
  console.log(parts[0]);
}```

File: common/models/middleware.js (hunk 2 above)
Comment (deepakrkris) [inline common/models/middleware.js]: agreed

File: common/models/middleware.js (hunk 4 above)
Comment (deepakrkris) [inline common/models/middleware.js]: agreed

File: component/datamodel/middleware-phase.js (hunk 5 above)
Comment (deepakrkris) [inline component/datamodel/middleware-phase.js]: agreed

File: component/datamodel/util/file-utility.js (hunk 6 above)
Comment (deepakrkris) [inline component/datamodel/util/file-utility.js]: yes, agreed, the callback should be called

File: component/tasks.js (hunk 7 above)
Comment (deepakrkris) [inline component/tasks.js]: tasks.js is a mixin class for workspace.js. I just want to be clear here so that other developers will understand that.

File: component/workspace.js (hunk 8 above)
Comment (deepakrkris) [inline component/workspace.js]: it is used below in ```
      let phase = workspace.getMiddlewarePhase(phaseName);
```

connector/index.js
deepakrkris: the callback is passed here as 'cb'

connector/middleware-handler.js
deepakrkris: agreed

common/models/middleware.js
eddiemonge: I meant if id isn't a string it will throw an error so split should always return an array and arrays always have a length property

component/workspace.js
eddiemonge: I still don't see `phases` in that string

deepakrkris: that is right, phases is unused. will remove that

common/models/middleware.js
deepakrkris: agreed

PR #388, opened 2017-01-11
Read workspace artifacts

deepakrkris:
Add loopback methods to read workspace artifacts

@superkhau PTAL


superkhau:
This was merged via a git knowledge share pair programming session, was good to merge.

PR #408, opened 2017-01-17
Fix CI errors in end to end test

deepakrkris:
- [X] - require sandbox app and mount rest api root
- [X] -  Skip failing tests.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+2017-01-23, Version 3.37.0
+==========================
+
+ * Enable "logoutSessionsOnSensitiveChanges" (Miroslav Bajtoš)
+
+ * test cleanup: extract bootSandboxWithOptions() (Miroslav Bajtoš)
+
+
 2017-01-05, Version 3.36.1
 ==========================
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "loopback-workspace",
-  "version": "3.36.1",
+  "version": "3.37.0",
   "main": "server/server.js",
   "publishConfig": {
     "export-tests": true


deepakrkris:
```
  end-to-end hello-world template
loadApp
[ 'User', 'AccessToken', 'ACL', 'RoleMapping', 'Role', 'Message' ]
{ 'Message.greet': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.prototype.__findById__accessTokens': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.prototype.__destroyById__accessTokens': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [],
     errors: [] },
  'User.prototype.__updateById__accessTokens': 
   { http: undefined,
     accepts: [ [Object], [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.prototype.__get__accessTokens': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.prototype.__create__accessTokens': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.prototype.__delete__accessTokens': { http: undefined, accepts: [], returns: [], errors: [] },
  'User.prototype.__count__accessTokens': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.create': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.upsert': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.exists': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.findById': 
   { http: undefined,
     accepts: [ [Object], [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.find': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.findOne': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.updateAll': 
   { http: undefined,
     accepts: [ [Object], [Object] ],
     returns: [],
     errors: [] },
  'User.deleteById': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [],
     errors: [] },
  'User.count': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.prototype.updateAttributes': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.createChangeStream': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.login': 
   { http: undefined,
     accepts: [ [Object], [Object] ],
     returns: [ [Object] ],
     errors: [] },
  'User.logout': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [],
     errors: [] },
  'User.confirm': 
   { http: undefined,
     accepts: [ [Object], [Object], [Object] ],
     returns: [],
     errors: [] },
  'User.resetPassword': 
   { http: undefined,
     accepts: [ [Object] ],
     returns: [],
     errors: [] } }
Unhandled error for request GET /api/Users: Error: Cannot GET /api/Users
    at raiseUrlNotFoundError (C:\jenkins\workspace\loopback-workspace\d60c4a15\test\sandbox\node_modules\loopback\server\middleware\url-not-found.js:15:17)
```

deepakrkris:
Printing app._router.stack gives the following , but /api/Users returns url not found
```[ { handle: [Function: query],
    name: 'query',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: { /^\/?(?=\/|$)/i fast_slash: true },
    route: undefined,
    phase: { builtin: true } },
  { handle: [Function: expressInit],
    name: 'expressInit',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: { /^\/?(?=\/|$)/i fast_slash: true },
    route: undefined,
    phase: { builtin: true } },
  { handle: [Function: favicon],
    name: 'favicon',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial:before' },
  { handle: [Function: compression],
    name: 'compression',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function],
    name: '<anonymous>',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function: xXssProtection],
    name: 'xXssProtection',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function: frameguard],
    name: 'frameguard',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function: hsts],
    name: 'hsts',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function: hidePoweredBy],
    name: 'hidePoweredBy',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function: ienoopen],
    name: 'ienoopen',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: [Function: nosniff],
    name: 'nosniff',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'initial' },
  { handle: 
     { [Function: router]
       params: {},
       _params: [],
       caseSensitive: undefined,
       mergeParams: undefined,
       strict: undefined,
       stack: [Object] },
    name: 'router',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /^\/explorer\/?(?=\/|$)/i,
    route: undefined },
  { handle: 
     { [Function: router]
       params: {},
       _params: [],
       caseSensitive: undefined,
       mergeParams: undefined,
       strict: undefined,
       stack: [Object] },
    name: 'router',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: { /^\/?(?=\/|$)/i fast_slash: true },
    route: undefined },
  { handle: [Function: restApiHandler],
    name: 'restApiHandler',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:^${restApiRoot}\/?(?=\/|$))/i,
    route: undefined,
    phase: 'routes' },
  { handle: [Function: raiseUrlNotFoundError],
    name: 'raiseUrlNotFoundError',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'final' },
  { handle: [Function: strongErrorHandler],
    name: 'strongErrorHandler',
    params: undefined,
    path: undefined,
    keys: [],
    regexp: /(?:)/i,
    route: undefined,
    phase: 'final:after' } ]```

File: test/end-to-end.js [hunk 1]
```diff
@@ -1185,3 +1189,9 @@ function readBuiltinPhasesFromSanbox() {
   app.lazyrouter(); // initialize request handling phases
   return app._requestHandlingPhases;
 }
+
+function getSandBoxApp() {
+  var app = require(SANDBOX);
+  app.use(app.get('restApiRoot'), app.loopback.rest());
```
bajtos: I am afraid this is removing an important point of these end-to-end tests, which is to verify the behaviour of the application scaffolded by loopback-workspace. Also note that this is registering the middleware after all other middleware from `server/middleware.json` has been already registered. The order in which the middleware is registered is important, as Express checks the middleware one by one until some handler matches. If the app was scaffolded and bootstrapped correctly, then this line would be a no-op because all loopback applications add `urlNotFound` handler that converts all requests to a 404 error.

I suspect that `require(SANDBOX)` is not correctly booting up the application.

Most likely, the bootstrapping process is asynchronous and you need to wait for `app.once('booted')` before continuing with the test.

test/end-to-end.js
```diff
@@ -146,8 +146,8 @@ describe('end-to-end', function() {
         .expect(404, done);
     });
 
-    it('comes with loopback-component-explorer', function(done) {
-      request(app).get('/explorer/swagger.json')
+    it.skip('comes with loopback-component-explorer', function(done) {
```
bajtos: Please add a `TODO` comment pointing to the follow-up issue, one comment per each `it.skip` please.

test/end-to-end.js
```diff
@@ -146,8 +146,8 @@ describe('end-to-end', function() {
...
+      request(app).get('/api/explorer/swagger.json')
```
bajtos: Please revert, explorer is always mounted at `/explorer/swagger.json`.

test/cachedInstall.js
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Install dependencies of the package in `appDir` directory.
+ * @param {string} appDir Application's root directory.
+ * @param {string} cacheDir Directory where to keep cached files.
+ * @param {Array<string>=} depTypes Optiona list of dependency types to install,
+ * e.g. `['dependencies', 'devDependencies']`. Defaults to `['dependencies']`.
+ * @param {function(Error=)} cb The callback.
+ */
+module.exports = function install(appDir, cacheDir, depTypes, cb) {
```
bajtos: Cached install is our module. If there is any fix needed to make loopback-workspace tests work, then please send a pull request to https://www.npmjs.com/package/strong-cached-install

We shouldn't be keeping an out-of-sync copy of that module in loopback-workspace.

test/end-to-end.js
```diff
@@ -960,7 +963,8 @@ describe('end-to-end', function() {
             .end(function(err) {
               if (err) return done(err);
               // localhost:3000 is the default value provided by loopback
-              expectAppIsRunning('http://localhost:3000', done);
+              //expectAppIsRunning('http://localhost:3000', done);
```
bajtos: We don't keep commented-out code. Please revert this change. If the test is failing, then it's better to temporarily skip it via `it.skip`.

superkhau:
@bajtos I took a look at this PR and @deepakrkris explained to me over hangout that the root cause of these failures is something in loopback-boot. Fixing that error will ultimately these set of tests. I am proposing we do:

- [ ] Add platform specific skips like https://github.com/strongloop/loopback-connector-mysql/blob/master/test/migration.test.js#L253 (he mentioned the tests fail due to the rest middleware not mounting properly on windows) -- leave all the workspace code alone like you mentioned in this PR's comments
- [ ] Skip all these workspace tests to get CI green AFTER creating follow up issues and inline comment/references to the issues (individual issues preferred so we can pull in 1 per sprint, etc and work on the fix incrementally over the next few sprints) -- all these issues should have major labels
- [ ] Speak to @rmg to figure out why the tests are being rerun on any fail (which @deepakrkris explained is another anomoly with the current CI)
- [ ] Create an issue in loopback-boot to figure out steps to reproduce/write failing test in loopback-boot to prove the middleware mount failing on windows (with major label)

Doing all this will get us to a green state in this repo now. Once we're in a steady state, I would say @ritch can start planning the other follow up issues based on current priorities during backlog grooming.

bajtos:
@superkhau thanks, your proposal makes sense.

> Creating follow up issues (individual issues preferred so we can pull in 1 or 2 per sprint to incrementally fix the problem)

Is this actually needed? If the problem is in loopback-boot, then we should fix that part first and the see what are the remaining issue in workspace. I'd like to avoid short-term fixes in loopback-workspace that will become obsolete once we fix loopback-boot. (Adding platform-specific `skip` is an acceptable short-term solution.)

>  platform specific skips like platform specific skips like https://github.com/strongloop/loopback-connector-mysql/blob/master/test/migration.test.js#L253 

The way how the skip is implemented there is pretty bad, as it does not provide any indication about the fact that a test was skipped, and thus creates a false sense of security/safety.

We should define a helper, e.g. `it.skipIf`, that will call either `it` or `it.skip` depending on the condition.

```js
it.skipIf = function(cond, name, fn) {
  if (cond)
    it.skip(name);
  else
    it(name, fn);
};

// in the test
it.skipIf(platform.isWindows, 'should autoupdate', function(done) {
  // ...
}
```

superkhau:
>Is this actually needed? If the problem is in loopback-boot, then we should fix that part first and the see what are the remaining issue in workspace. I'd like to avoid short-term fixes in loopback-workspace that will become obsolete once we fix loopback-boot. (Adding platform-specific skip is an acceptable short-term solution.)

Yeah, that part is debatable -- I think fixing loopback-boot may ultimately resolve all the issues. I only put it there because @deepakrkris mentioned he wanted to take on each of those issues incrementally. 

>The way how the skip is implemented there is pretty bad, as it does not provide any indication about the fact that a test was skipped, and thus creates a false sense of security/safety.

Yeah, it was a short term fix I proposed to @loay. Your solution with the helper is better -- same as we do in kv connector capabilities (ie. the current solution in the MySQL connector is pretty much a false positive when I look back at it now :see_no_evil:). @loay I created https://github.com/strongloop/loopback-connector-mysql/issues/234, please get that tech debt planned as your discretion.

test/end-to-end.js
```diff
@@ -29,6 +30,14 @@ var MYSQL_DATABASE = 'loopback_workspace_test';
 var MYSQL_USER = 'lbws';
 var MYSQL_PASSWORD = 'hbx42rec';
 
+// add conditional iterations
+it.skipIf = function(cond, name, fn) {
+  if (cond)
+    it.skip(name);
+  else
+    it(name, fn);
+};
```
superkhau: move to bdd helper like https://github.com/strongloop/loopback-datasource-juggler/blob/master/test/helpers/bdd-if.js

test/end-to-end.js
```diff
@@ -146,8 +155,9 @@ describe('end-to-end', function() {
         .expect(404, done);
     });
 
-    it('comes with loopback-component-explorer', function(done) {
-      request(app).get('/explorer/swagger.json')
+    // TODO fix the skipped test https://github.com/strongloop/loopback-workspace/issues/411
```
superkhau: TODO follow up issue for fix at ...

test/end-to-end.js
```diff
@@ -1185,3 +1204,9 @@ function readBuiltinPhasesFromSanbox() {
   app.lazyrouter(); // initialize request handling phases
   return app._requestHandlingPhases;
 }
+
+function getSandBoxApp() {
+  var app = require(SANDBOX);
+  app.use(app.get('restApiRoot'), app.loopback.rest());
```
superkhau: if (platform.isWindows)

superkhau:
Follow up issues at:

- https://github.com/strongloop/loopback-workspace/issues/411
- https://github.com/strongloop/loopback-workspace/issues/412
- https://github.com/strongloop/loopback-workspace/issues/413
- https://github.com/strongloop/loopback-workspace/issues/414

test/end-to-end.js
```diff
@@ -163,6 +173,7 @@ describe('end-to-end', function() {
       };
       boot(app, bootOptions, function(err) {
         if (err) return done(err);
+        app.use(app.get('restApiRoot'), app.loopback.rest());
```
bajtos: Please remove, the `REST `middleware must be added by `loopback-boot`. If it is not, then skip the test for now.

test/end-to-end.js
```diff
@@ -214,7 +228,6 @@ describe('end-to-end', function() {
       };
       boot(app, bootOptions, function(err) {
         if (err) return done(err);
-
```
bajtos: Whitespace-only change, please revert

test/end-to-end.js
```diff
@@ -961,6 +977,7 @@ describe('end-to-end', function() {
               if (err) return done(err);
               // localhost:3000 is the default value provided by loopback
               expectAppIsRunning('http://localhost:3000', done);
+              done();
```
bajtos: `expectAppIsRunning` is already accepting `done` callback. Please revert this change and fix `expectAppIsRunning` if needed.

test/end-to-end.js
```diff
@@ -1185,3 +1204,9 @@ function readBuiltinPhasesFromSanbox() {
   app.lazyrouter(); // initialize request handling phases
   return app._requestHandlingPhases;
 }
+
+function getSandBoxApp() {
```
bajtos: Nitpick: perhaps `loadSandBoxApp` would be a better name?

test/end-to-end.js
bajtos: No no, the REST adapter must be added by loopback-boot. If it isn't, then we should skip the test for now.

Please remove this line.

```js
function loadSandboxApp() {
  var app = require(SANDBOX);
  // TODO wait until the app boots
  return app;
}
```

test/end-to-end.js
```diff
@@ -19,6 +19,7 @@ var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
 var should = require('chai').should();
+var platform = require('./helpers/platform');
```
bajtos: Perhaps `var isWindows = require('./helpers/platform').isWindows` to make the test code below shorter and more likely to fit within 80 chars?

test/end-to-end.js
superkhau: Yes, he is using that line as a hack to load it on windows. But yes, I can also agree with skipping the entire test for now since boot isn't loading it. I'm ok with:

1. load only for windows tests
2. skip the test altogether (as you suggested)

test/end-to-end.js
superkhau: We really should just getApp or whatever (the sandboxing is part of the impl and doesn't need to be explicit in the name at all IMO).

test/end-to-end.js
```diff
@@ -1215,15 +1240,8 @@ function readBuiltinPhasesFromSanbox() {
   return app._requestHandlingPhases;
 }
 
-function bootSandboxWithOptions(options, done) {
```
bajtos: I think this was removed after rebase, please bring it back.

test/end-to-end.js
```diff
@@ -171,7 +182,12 @@ describe('end-to-end', function() {
       });
     });
 
-    it('omits sensitive error details in production mode', function(done) {
+    // TODO fix skipped tests https://github.com/strongloop/loopback-workspace/issues/412
+    it.skipIf(platform.isWindows, 'omits sensitive error details in production mode',
+    function(done) {
+      var loopback = require(SANDBOX + '/node_modules/loopback');
+      var boot = require(SANDBOX + '/node_modules/loopback-boot');
+      var app = loopback({ localRegistry: true, loadBuiltinModels: true });
```
bajtos: These three lines are no longer needed, see `bootSandboxWithOptions`

test/end-to-end.js
```diff
@@ -194,15 +210,18 @@ describe('end-to-end', function() {
       });
     });
 
-    it('disables built-in REST error handler', function(done) {
+    // TODO fix skipped tests https://github.com/strongloop/loopback-workspace/issues/412
+    it.skipIf(platform.isWindows, 'disables built-in REST error handler', function(done) {
+      var loopback = require(SANDBOX + '/node_modules/loopback');
+      var boot = require(SANDBOX + '/node_modules/loopback-boot');
+      var app = loopback({ localRegistry: true, loadBuiltinModels: true });
```
bajtos: These three lines are no longer needed, see `bootSandboxWithOptions`

test/end-to-end.js
```diff
@@ -1215,15 +1240,8 @@ function readBuiltinPhasesFromSanbox() {
...
-  var bootOptions = extend({
-    appRootDir: SANDBOX + '/server',
-  }, options);
-
-  boot(app, bootOptions, function(err) {
-    done(err, app);
-  });
+function getSandBoxApp() {
+  var app = require(SANDBOX);
+  app.use(app.get('restApiRoot'), app.loopback.rest());
```
bajtos: As was discussed earlier, please remove this hack.

test/end-to-end.js
```diff
@@ -1215,15 +1240,8 @@ function readBuiltinPhasesFromSanbox() {
...
-  var boot = require(SANDBOX + '/node_modules/loopback-boot');
-  var app = loopback({ localRegistry: true, loadBuiltinModels: true });
-  var bootOptions = extend({
-    appRootDir: SANDBOX + '/server',
-  }, options);
-
-  boot(app, bootOptions, function(err) {
-    done(err, app);
-  });
+function getSandBoxApp() {
```
bajtos: @superkhau proposed in an earlier comment to drop `sandbox` from the name. I am proposing to replace `get` with `load`. That would lead us to `loadApp` or perhaps `loadScaffoldedApp` to make it more clear what kind of application we are loading.

bajtos:
Please wait with further work on this front until https://github.com/strongloop/strong-cached-install/pull/13 is landed and published. I think it may fix most if not all of the Windows failures.

bajtos:
strongloop/strong-cached-install#13 was landed and released, I think we can revert most of the changes in this patch now.

deepakrkris:
no changes seems to be required

PR #418, opened 2017-01-25
Fix CI Errors in windows

deepakrkris:
Fix Windows CI Errors:

- [x] - remove cached install
- [x] - add 'npm install' in sandbox
- [x] - initialize sandbox with 'node_modules' folder 
- [x] - loopback-workspace should use latest loopback-boot

### DIFF
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -7,7 +7,6 @@ var async = require('async');
 var exec = require('child_process').exec;
 var extend = require('util')._extend;
 var fs = require('fs-extra');
-var install = require('strong-cached-install');
 var mysql = require('mysql');
 var path = require('path');
 var request = require('supertest');
@@ -18,6 +17,7 @@ var models = workspace.models;
 var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
+var platform = require('./helpers/platform');
 var should = require('chai').should();
 
 var Workspace = workspace.models.Workspace;
@@ -28,6 +28,14 @@ var PKG_CACHE = path.resolve(__dirname, '.pkgcache');
 var MYSQL_DATABASE = 'loopback_workspace_test';
 var MYSQL_USER = 'lbws';
 var MYSQL_PASSWORD = 'hbx42rec';
+
+it.skipIf = function(condition, desc, fn) {
+  if (condition) {
+    it.skip(desc, fn);
+  } else {
+    it(desc, fn);
+  }
+};
 
 describe('end-to-end', function() {
   this.timeout(15000);
@@ -704,7 +712,8 @@ describe('end-to-end', function() {
       });
     });
 
-    it('updates all models in the database', function(done) {
+    it.skipIf(platform.isWindows, 'updates all models in the database',
+    function(done) {
       db.autoupdate(undefined, function(err) {
         if (err) return done(err);
         listTableNames(connection, function(err, tables) {
@@ -1179,7 +1188,8 @@ function execNpm(args, options, cb) {
 
 function installSandboxPackages(cb) {
   this.timeout(300 * 1000);
-  install(SANDBOX, PKG_CACHE, ['dependencies', 'devDependencies'], cb);
+  initializePackage();
+  localInstall(SANDBOX, cb);
 }
 
 function listTableNames(connection, cb) {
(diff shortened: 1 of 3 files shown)


rmg:
@slnode test please

deepakrkris:
@slnode test please

deepakrkris:
@slnode test please

bajtos:
@slnode test please

bajtos:
We may need to increase the timeout in end-to-end tests even more.

See http://ci.strongloop.com/job/loopback-workspace/node=4.x,os=windows/6113/console

```
 end-to-end empty-server template
      1) "before all" hook: loadApp

  Suite duration: 110.692 s, Tests: 0

  end-to-end empty-server template without explorer
      1) "before all" hook: loadApp

  Suite duration: 35.175 s, Tests: 0

  end-to-end api-server template
      1) "before all" hook: loadApp

  Suite duration: 45.529 s, Tests: 0
```

deepakrkris:
@bajtos , booting of the app in windows works fine. it is the route configuration in middleware.json (placed in templates folder) that didnt get configured. When I added a boot script to the templates, the issue was resolved. I will be removing the file later. 
@bajtos @superkhau  I am debugging loopback-boot in windows now.

bajtos:
> booting of the app in windows works fine. it is the route configuration in middleware.json (placed in templates folder) that didnt get configured. When I added a boot script to the templates, the issue was resolved. I will be removing the file later. I am debugging loopback-boot for the issue in windows now.

Thank you for the update. The two tests you are disabling in this patch were passing for me on my local Windows machine. I think you need to delete `test/.pkgcache` after you installed the latest version of strong-cached-install to force strong-cached-install to download newer (the latest version) of all dependencies. Without that step, you will keep using the cached files fetched via `npm install loopback-boot@2.6.5` (a command with missing `^` in the version specifier).

Let me know if you need any more help. I can debug this myself to prevent too much waste of your time, if you like.

deepakrkris:
@bajtos @superkhau  I have made some more progress on the workspace CI issues in windows. I removed the additional boot script and changed the “routes” middleware configuration from `${restApiRoot}` to `/api` and all tests are clear (except for the 8 I have skipped)

superkhau:
Good, let's keep going down this path. I'm guessing eventually this whole PR will be moot if you fix the other root causes.

deepakrkris:
@bajtos you are right. this is a version error. I see this functionality is missing in the installed boot version : https://github.com/strongloop/loopback-boot/pull/149

deepakrkris:
@bajtos @superkhau  I removed cached install and just did a plain 'npm install' and all tests are passing now (except for the skipped 6)

bajtos:
I introduced strong-cached-install in order to reduce the time needed to run end-to-end, so that we can get fast feedback when developing in TDD red-green-refactor style. Strangely enough, on Node v6.x using `npm@3` which produces flat layout in `node_modules`, strong-cached-install is slower than `npm install`. (`mocha test/end-to-end.js` takes 5 minutes on master, 4 minutes on this branch). In that light, I am fine with removing strong-cached-install.

test/end-to-end.js
```diff
@@ -689,7 +690,7 @@ describe('end-to-end', function() {
       });
     });
 
-    it('updates all models in the database', function(done) {
+    it.skip('updates all models in the database', function(done) {
```
bajtos: Why is this needed? Where is the follow-up issue to fix the tests on Windows properly? Please add a comment to the test file.

test/end-to-end.js
```diff
@@ -1227,3 +1229,15 @@ function bootSandboxWithOptions(options, done) {
     done(err, app);
   });
 }
+
+function localInstall(cwd, cb) {
+  var options = {
+    cwd: cwd,
+  };
+  var script = 'npm install';
+  return exec(script, options, function(err, stdout, stderr) {
```
bajtos: Please `debug` log the script and the options or at `cwd` before calling `exec`, e.g.

```js
debug('Running `%s` in %s', script, cwd);
exec(...);
```

test/helpers/platform.js
```diff
@@ -0,0 +1,8 @@
+// Copyright IBM Corp. 2013,2016. All Rights Reserved.
+// Node module: loopback-connector-mysql
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+
+'use strict';
+
+exports.isWindows = /^win/.test(process.platform);
```
bajtos: I believe this helper is no longer used, please remove.

test/support.js
```diff
@@ -84,6 +84,10 @@ givenFile = function(name, pathToFile) {
   };
 };
 
+initializePackage = function() {
```
bajtos: Since this helper is used only by `installSandboxPackages`, please move its definition the `test/end-to-end.js`.

test/support.js
```diff
@@ -84,6 +84,10 @@ givenFile = function(name, pathToFile) {
...
+  fs.mkdirSync(path.join(SANDBOX, 'node_modules'));
```
bajtos: Please add a comment explaining why is this needed.

test/end-to-end.js
```diff
@@ -18,6 +17,7 @@ var models = workspace.models;
 var TestDataBuilder = require('./helpers/test-data-builder');
 var ref = TestDataBuilder.ref;
 var given = require('./helpers/given');
+var platform = require('./helpers/platform');
```
bajtos: AFAICT, this is not used, please remove.

test/helpers/platform.js
deepakrkris: used to skip one test

test/end-to-end.js
superkhau: Nevermind, i see the comment below where you need it to skip a test

PR #419, opened 2017-01-25
Additional templates to create workspace

deepakrkris:
Add more available templates and extension

- [x] -  allow template extension
- [x] - add template files to copy
- [x] - add more templates


templates/config/api-server.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "api-server",  
+  "description": "A LoopBack API server with local User auth",
+  "extends": "empty-server",
+  "files": {
+    "path": "api-server"
```
superkhau: ./api-server

PR #427, opened 2017-02-03
4.x : Add end to end tests

deepakrkris:
Fix workspace models to run end to end test on scaffolded application

- [ ] - split Id using ':' in Middleware model
- [ ] - Add models metadata to model config
- [ ] - load model config
- [ ] - fix server.js in template
- [ ] - add end to end test to require created api-server (more tests will be added later)


package.json
```diff
@@ -8,7 +8,9 @@
   "scripts": {
     "lint": "eslint .",
     "pretest": "node test/helpers/pretest.js",
-    "test": "grunt",
```
bajtos: Assuming integration tests run faster, then I think the order should be different - run the fast tests first to get quick feedback, run the slower acceptance suite only after we verified that the building blocks work correctly.

templates/files/empty-server/server/middleware.development.json
```diff
@@ -1,10 +0,0 @@
-{
-  "final:after": {
```
bajtos: IIUC, you are changing the template of `empty-server` here, compared to what the current loopback-workspace@3 scaffolds.

What's the reasoning behind this change?

.gitignore
```diff
@@ -1,3 +1,2 @@
 node_modules
 test/sandbox
-test/sandbox2
```
bajtos: Appending a number to directory name does not scale well. If you have a test that needs multiple directories, then you should use nested subdirectories inside `test/sandbox`, e.g. `test/sandbox/app1` and `test/sandbox/app2`.

component/datamodel/facet.js
```diff
@@ -36,23 +37,40 @@ class Facet extends Entity {
...
+    Object.keys(config).forEach(function(key) {
+      if (key === '_meta') return;
+      let modelConfig = modelConfigNodes[key];
+      if (modelConfig) {
+        modelConfig._content = config[key];
+      } else {
+        if (workspace.getModel(key)) {
+          this.addModelConfig(workspace, key, config[key]);
+        } else
+        this.addModelConfig(workspace, 'common.models.' + key, config[key]);
```
superkhau: let chosenKey = workspace.getModel(key) ? key : 'common.models';

addModelConfig(workspac, chosenKey, config[key]):

test/helpers/test-support.js
```diff
@@ -23,3 +28,24 @@ function givenEmptySandbox(sandboxDir, cb) {
...
+    localInstall(dir, cb);
+  });
+}
+
+function localInstall(cwd, cb) {
+  var options = {
+    cwd: cwd,
+  };
+  var script = 'npm install';
+  return exec(script, options, function(err) {
```
superkhau: return exec(script, options, cb);

superkhau: return exec('npm install', options, cb);

package.json
deepakrkris: agreed, discussed with Simon, the current integration tests are actually end to end tests and will be moved to acceptance tests in the next PR. I will then add unit and integration tests.

templates/files/empty-server/server/middleware.development.json
deepakrkris: I havent yet added the capability to read middleware.development.json. I will add the middleware.development.json in a future PR.

.gitignore
deepakrkris: I am actually removing the test/sandbox2 folder here

PR #428, opened 2017-02-06
4.x : Template inheritance

deepakrkris:
- [ ] - Extend the parent template with child templates 
 - [ ] - create a tree to link templates in the order of inheritance


templates/files/hello-world/common/models/message.json
```diff
@@ -2,10 +2,12 @@
   "name": "Message",
   "facetName": "common",
   "properties": {},
+  "readonly": true,
+  "strict": true,
+  "public": true,
   "idInjection": true,
   "methods": {
     "greet": {
-      "isStatic": true,
```
superkhau: add back

deepakrkris: agreed

PR #433, opened 2017-02-14
4.x : Add functionality to create acl

deepakrkris:
Add access control list for loopback model

### DIFF
--- a/test/acceptance/use-cases/case-03-test-multiple-workspaces.feature
+++ b/test/acceptance/use-cases/case-03-test-multiple-workspaces.feature
@@ -48,6 +48,10 @@ Feature: Users should be able to create middleware configurations
     When I add property 'name' of type 'string'
     Then the model property is created
 
+  Scenario: Add access control
+    When I add acl to model 'Customer'
+    Then the model acl is created
+
   Scenario: Add a custom method to the model
     Given I add model method 'addReview' in workspace 'api-server'
     When the method has an argument 'description' type 'string'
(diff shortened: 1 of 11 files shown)


test/acceptance/use-cases/test-scripts/create-acl.js
```diff
@@ -0,0 +1,42 @@
...
+      index: 0,
+      modelId: 'common.models.' + modelName,
+      property: 'find',
+      accessType: 'EXECUTE',
+      principalType: 'ROLE',
+      principalId: '$authenticated',
+      permission: 'ALLOW',
+    };
+    const options = {workspaceId: testsuite.workspace.getId()};
+    ModelAccessControl.create(config, options, function(err, data) {
```
superkhau: ModelAccessControl.create(config, options, next)

test/acceptance/use-cases/test-scripts/create-acl.js
```diff
@@ -0,0 +1,42 @@
...
+const ModelAccessControl = app.models.ModelAccessControl;
+
+app.on('booted', function() {
+  app.emit('ready');
+});
+
+module.exports = function() {
+  const testsuite = this;
+
+  this.When(/^I add acl to model '(.+)'$/,
```
superkhau: Given
  - an acl with these configs
    - with what role user (ROLE)
    - with permission set to ALLOW or DENY (the ACL)
    - operation execute on find
AND:
  - we have a user that is not logged (authenticated or unauthenticated)
When I execute the find operation on a model with

test/acceptance/use-cases/case-03-test-multiple-workspaces.feature
```diff
@@ -48,6 +48,10 @@ Feature: Users should be able to create middleware configurations
     When I add property 'name' of type 'string'
     Then the model property is created
 
+  Scenario: Add access control
+    When I add acl to model 'Customer'
+    Then the model acl is created
```
superkhau: data table

PR #437, opened 2017-02-17
Enable strictObjectIDCoercion for RoleMapping model

bajtos:
### Description

Force the RoleMapping model to always store ids as ObjectId when using MongoDB connector. This is needed to allow users to query RoleMapping via "principalId" property.

The solution uses a less-known feature where `server/model-config.json` can provide additional "options" to apply even to built-in models like `RoleMapping`.

#### Related issues

- https://github.com/strongloop/loopback/issues/3121
- https://github.com/strongloop/loopback/issues/1441
- https://github.com/strongloop/loopback/pull/3198 

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style

### DIFF
--- a/templates/projects/api-server/data.js
+++ b/templates/projects/api-server/data.js
@@ -43,6 +43,9 @@ template.server = {
     {
       name: 'RoleMapping',
       dataSource: 'db',
+      options: {
+        strictObjectIDCoercion: true,
+      },
       public: false,
     },
     {
--- a/templates/projects/hello-world/data.js
+++ b/templates/projects/hello-world/data.js
@@ -40,6 +40,9 @@ template.server = {
     {
       name: 'RoleMapping',
       dataSource: 'db',
+      options: {
+        strictObjectIDCoercion: true,
+      },
       public: false,
     },
     {
--- a/templates/projects/notes/data.js
+++ b/templates/projects/notes/data.js
@@ -40,6 +40,9 @@ template.server = {
     {
       name: 'RoleMapping',
       dataSource: 'db',
+      options: {
+        strictObjectIDCoercion: true,
+      },
       public: false,
     },
     {
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -394,6 +394,11 @@ describe('end-to-end', function() {
         .expect(422)
         .end(done);
     });
+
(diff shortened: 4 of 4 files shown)


bajtos:
@slnode test please

bajtos:
Released in `loopback-workspace@3.38.0`.

ebarault:
@bajtos : i just helped someone else for who ACLs were not working because RoleMapping had principalId as string, not ObjectId. This bug is painful.

Remind me please: why can't we just add this option to the built-in RoleMapping model in loopback core (or do the config in app.js at app boot if a feature flag is required) ?

```js    
"options": {
      "strictObjectIDCoercion": true
    },
```

bajtos:
@ebarault 

> Remind me please: why can't we just add this option to the built-in RoleMapping model in loopback core 

Backwards compatibility, as usually. See https://github.com/strongloop/loopback/pull/3198#issuecomment-280382924

> `strictObjectIDCoercion` will also dictate how ID's are written to mongo in the first place. In other words, it can only be enabled for 'fresh' projects, or after a db migration of legacy (string) ID's in existing applications.

You are welcome :)

---

> (or do the config in app.js at app boot if a feature flag is required) ?

I think that's what this pull request is effectively doing in new applications - setting the required feature flag the way how it can/should be set.

Existing applications can apply the same approach - simply modify the `RoleMapping` section in `server/model-config.json` to the following:

```json
{
  "RoleMapping": {
    "dataSource": "db",
    "options": {
      "strictObjectIDCoercion": true,
    },
    "public": false
  }
}
```

I do realise the current situation is unfortunate and I am happy to discuss other/better ways how to solve this problem.

I think ideally we should fix the root cause, which IIRC lies in the way how foreign keys are treated by LoopBack - the type of foreign keys should be changed from `any` to `number`, `string`, or `ObjectID` depending on what is the database the target model is attached to.

@superkhau was already taking a stab at that issue, but with little success. You can search for "fix ObjectID coercion", here are few issues to get started:
 - proposal for long-term: https://github.com/strongloop/loopback/issues/1874
 - short-term fix that was rejected: https://github.com/strongloop/loopback-datasource-juggler/pull/778

PR #460, opened 2017-03-02
Add integration test for facets

deepakrkris:
create new facet and check configurations

### DIFF
--- a/test/integration/facet.js
+++ b/test/integration/facet.js
@@ -1,0 +1,76 @@
+// Copyright IBM Corp. 2014,2016. All Rights Reserved.
+// Node module: loopback-workspace
+// This file is licensed under the MIT License.
+// License text available at https://opensource.org/licenses/MIT
+'use strict';
+
+const async = require('async');
+const app = require('../../');
+const expect = require('../helpers/expect');
+const fs = require('fs-extra');
+const testSupport = require('../helpers/test-support');
+const loopback = require('loopback');
+const Facet = app.models.Facet;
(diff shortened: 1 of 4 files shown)


test/integration/facet.js
```diff
@@ -0,0 +1,72 @@
...
+describe('Facet', function() {
+  describe('CRUD', function() {
+    let workspace, dir;
+    before(function(done) {
+      testSupport.givenBasicWorkspace('empty-server', function() {
+        workspace = WorkspaceManager.getWorkspace();
+        dir = workspace.getDirectory();
+        done();
+      });
+    });
```
superkhau: see style guide test layout

test/integration/facet.js
```diff
@@ -0,0 +1,72 @@
...
+
+    it('model.create()', function(done) {
+      Facet.create({
+        name: 'foo',
+      }, function(err, def) {
+        expect(err).to.not.exist();
```
superkhau: if (err) return done (err);

test/integration/facet.js
```diff
@@ -0,0 +1,72 @@
...
+        mixins: [
+          'loopback/common/mixins',
+          'loopback/server/mixins',
+          '../common/mixins',
+          './mixins',
+        ],
+      });
+    });
+
+    it('saves facet models to correct file', function() {
```
superkhau: to the

test/integration/facet.js
superkhau: http://loopback.io/doc/en/contrib/style-guide.html#layout-of-test-files -- see the bad section

PR #473, opened 2017-03-22
Add engines.node: >=4 to scaffolded projects

bajtos:
### Description

Add `engines: { node: '>=4' }` to scaffolded projects

Also skip running `npm test` in the scaffolded project when
Node.js version if < 4 (typically when running on CI).

#### Related issues

This is a prerequisite for #467 and #472

### Checklist

- [ ] Code conforms with the [style

### DIFF
--- a/templates/projects/empty-server/data.js
+++ b/templates/projects/empty-server/data.js
@@ -12,6 +12,9 @@ template.supportedLBVersions = ['2.x', '3.x'];
 template.package = {
   'version': '1.0.0',
   'main': 'server/server.js',
+  'engines': {
+    'node': '>=4',
+  },
   'scripts': {
     'lint': 'eslint .',
     'start': 'node .',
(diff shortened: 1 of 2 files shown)


rmg:
The main impact of this, I think, will be if/when the generated apps get deployed to a PaaS like Heroku or Bluemix that uses the `engines.node` property to determine the version of node to run.

I think this is a good change since more and more dependencies are requiring node >=4 and this helps make it more clear that older versions of node should not be used.

kraman:
👍

bajtos:
> loopback-workspace/node=0.12,os=windows — Failed! (7e216e0)

Looks like intermittent failure, the second re-run with DEBUG enabled has passed.

> [cis-jenkins] downstream: apim-ui@5.x — Failed! (7e216e0)
> [cis-jenkins] downstream: apim-ui@5.0.0.x — Failed! (7e216e0)

These two builds are failing because of an unrelated error - `npm install` fails with `peerinvalid` error caused by angular/angular-material version mismatch.

PR #480, opened 2017-03-27
4.x : get function for child nodes

deepakrkris:
composite parent with getter for children


lib/datamodel/graph/collection.js
```diff
@@ -0,0 +1,55 @@
...
+  get array() {
+    const def = [];
+    const nodes = this.nodes;
+    Object.keys(nodes).forEach(function(key) {
+      const node = nodes[key];
+      def.push(node);
+    });
+    return def;
+  }
+  contents(options) {
```
superkhau: toJson()

lib/datamodel/graph/collection.js
```diff
@@ -0,0 +1,55 @@
...
+      const content = clone(node._content);
+      if (options && options.filter && options.filter.length)
+        options.filter.forEach(function(fieldName) {
+          delete content[fieldName];
+        });
+      def[key] = content;
+    });
+    return def;
+  }
+  get values() {
```
superkhau: get nodeContents

lib/datamodel/graph/collection.js
```diff
@@ -0,0 +1,55 @@
...
+class Collection {
+  constructor(nodes) {
+    this.nodes = {};
+    if (nodes)
+      this.nodes = nodes;
+  }
+  push(node) {
+    this.nodes[node._name] = node;
+  }
+  get array() {
```
superkhau: toArray()

superkhau:
```
get toArray() {
  return iterate(function filter(node) {
    return node;
  });
}

function iterate() {
  ... 
}
```

PR #482, opened 2017-03-28
fix windows perms fail

anthonyettinger:
add strict flag for config-file

- ignore permission issues when looking for config files
- fixes windows and mac issues

### DIFF
--- a/common/models/config-file.js
+++ b/common/models/config-file.js
@@ -211,7 +211,8 @@ module.exports = function(ConfigFile) {
     });
 
     function find(pattern, cb) {
-      glob(pattern, { cwd: workspaceDir }, cb);
+      // set strict to false to avoid perm issues
+      glob(pattern, { cwd: workspaceDir, strict: false }, cb);
     }
   };
 


anthonyettinger:
@eddie it still prints a warning in the console.

common/models/config-file.js
```diff
@@ -211,7 +211,8 @@ module.exports = function(ConfigFile) {
     });
 
     function find(pattern, cb) {
-      glob(pattern, { cwd: workspaceDir }, cb);
+      // set strict to false to avoid perm issues
```
superkhau: perm issues on windows (7?)

superkhau:
Please update the comment, otherwise LGTM.

common/models/config-file.js
anthonyettinger: its an issue on all platforms -- but this arose on windows 10.

anthonyettinger:
@slnode test please

common/models/config-file.js
superkhau: icic, let's add that in the comment, otherwise LGTM

// set strict to false to avoid perm issues on windows 10, may affect all platforms

anthonyettinger:
@slnode test please

anthonyettinger:
peer dependency failing on angular 1.6, ignoring.

PR #490, opened 2017-06-20
Support IBM Object Storage

hacksparrow:
### Description

Adds IBM Object Storage to list of available connectors

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -30,6 +30,41 @@
     "supportedByStrongLoop": true
   },
   {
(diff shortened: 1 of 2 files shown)


available-connectors.json
```diff
@@ -30,6 +30,41 @@
     "supportedByStrongLoop": true
   },
   {
+    "name": "object-storage",
```
raymondfeng: Maybe it should be `ibm-object-storage`

PR #496, opened 2017-08-31
Add generateOperationScopedModels flag

rashmihunt:
@raymondfeng PTAL

Related PRs
- https://github.com/strongloop/loopback-swagger/pull/99
- https://github.com/strongloop/loopback-component-explorer/pull/219

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "loopback-workspace",
-  "version": "4.0.0",
+  "version": "4.0.0-alpha.1",
   "main": "server/server.js",
   "publishConfig": {
     "export-tests": true
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "loopback-workspace",
-  "version": "3.42.0",
+  "version": "4.0.0",
   "main": "server/server.js",
   "publishConfig": {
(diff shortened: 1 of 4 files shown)


templates/projects/empty-server/data.js
```diff
@@ -97,6 +97,7 @@ template.server = {
       name: 'loopback-component-explorer',
       value: {
         mountPath: '/explorer',
+        generateOperationScopedModels: true,
```
raymondfeng: Please add a comment about `generateOperationScopedModels`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "loopback-workspace",
-  "version": "3.42.0",
+  "version": "4.0.0",
```
bajtos: We usually use a pre-release version number in a patch like this (e.g. `4.0.0-alpha.1`) and then let `slt-release` change it to the "production" version number.

rashmihunt: @bajtos done PTAL

templates/projects/empty-server/data.js
rashmihunt: @raymondfeng done PTAL

rashmihunt:
@slnode test please

PR #497, opened 2017-10-13
[WIP] Update template with a newer version of helmet

dhmlau:
Suggested by @bajtos , updating the template with a newer version of `helmet`. 
> We should update our project template to install the latest helmet version too, see https://github.com/strongloop/loopback-workspace/blob/8726d7ce7662051fe3236c1174d132c8ab038261/templates/projects/empty-server/data.js#L26

I've also updated the min version for `debug`

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "loopback-workspace",
   "version": "4.0.0",
   "main": "server/server.js",
+  "engines": { 
+    "node": ">=4"
+  },
   "publishConfig": {
     "export-tests": true
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
   "dependencies": {
     "async": "^1.5.2",
     "cookie-parser": "^1.3.2",
-    "debug": "^2.2.0",
+    "debug": "^2.6.9",
```
bajtos: This is effectively a no-op, because `^2.2.0` will install the latest 2.x version, e.g. `2.6.9.` Did you perhaps mean `^3.1.0`?

dhmlau:
@slnode  test please

bajtos:
I suspect helmet changed the default behaviour or perhaps the configuration format between major versions and that's why our tests are failing after upgrade. I can look into this next week if you like - just let me know, @dhmlau.

dhmlau:
@bajtos , sorry, i've been juggling with a few things and haven't got the chance to get to it.  Will try to get back to it tmr.

PR #506, opened 2018-02-26
change available-connectors.json to an object

shimks:
### Description

Currently, our CLI tool has no way of skipping prompts for datasource selections. In order to do this, flags need to be used to bypass the prompts, but it is impractical to fetch a specific connector with current implementation of loading the controllers via an array. The PR changes the array implementation to that of a map, so that lookups for a specific connector could be faster.

### DIFF
--- a/available-connectors.json
+++ b/available-connectors.json
@@ -1,5 +1,5 @@
-[
-  {
+{
+  "memory": {
     "name": "memory",
     "description": "In-memory db",
     "baseModel": "PersistedModel",
@@ -19,7 +19,7 @@
     },
     "supportedByStrongLoop": true
   },
-  {
+  "kv-memory": {
     "name": "kv-memory",
     "description": "In-memory key-value connector",
     "baseModel": "KeyValueModel",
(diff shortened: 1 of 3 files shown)


available-connectors.json
```diff
@@ -29,7 +29,7 @@
     },
     "supportedByStrongLoop": true
   },
-  {
+  "object-storage": {
```
jannyHou: prefix `ibm-`?

shimks:
Tests are failing on master btw

available-connectors.json
shimks: i thought it was verbose to have `ibm-` in front, but i'll introduce it back since it might not be clear enough

test/workspace.js
```diff
@@ -162,6 +162,14 @@ describe('Workspace', function() {
     });
   });
 
+  describe.only('project.getConnectorByName', function() {
```
kjdelisle: You left .only on this!

available-connectors.json
```diff
@@ -735,7 +735,7 @@
     },
     "supportedByStrongLoop": false
   },
-  {
+  "zos": {
```
kjdelisle: `zos` is an OS. This is too generic, please make it something like `zosconnectee` or `zos-connect-ee`.

shimks:
@slnode test please

shimks:
@slnode test please

shimks:
@slnode test please

shimks:
@slnode test please

PR #523, opened 2018-07-24
Extract string for project template desc

dhmlau:
### Description
Fixes https://github.com/strongloop/generator-loopback/issues/328

When running `apic lb`, users are prompted with which template to choose.  According to @raymondfeng , the name should not be translated, however, the description of the template can be translated.  This PR is to externalize the description message.


dhmlau:
After the fix, we'll get this output when the locale has switched to chinese.  I've made up the translation string just for verification.  

I'm not sure if this PR can be merged with downstream `generator-loopback` CI failed.  It's failing due to linting error.  I tried to fix it because it requires lots of changes.  It involves updating the eslint-* depenencies and the changes are about fixing the linting error.  I'm good to submit another PR to fix it. Just thinking that it might be better to do it in a separate PR.  Thanks.

hacksparrow:
👍

PR #529, opened 2018-09-28
Update loopback version

jannyHou:
### Description

Update loopback versions.
5085 is using 3.x, and it needs to include the security fix in loopback@3.22.0

#### Related issues

- connect to <link_to_referenced_issue>

### Checklist

- [ ] Code conforms with the [style

### DIFF
--- a/common/models/workspace.js
+++ b/common/models/workspace.js
@@ -36,14 +36,15 @@ module.exports = function(Workspace) {
     var TEMPLATE_DIR = path.join(__dirname, '..', '..', 'templates', 'projects');
     var DEFAULT_TEMPLATE = 'api-server';
     var DEPENDENCIES_3_X = {
-      'loopback': '^3.0.0',
-      'loopback-component-explorer': '^4.0.0',
+      'loopback': '^3.22.0',
+      'loopback-component-explorer': '^6.0.0',
     };
     var DEPENDENCIES_2_X = {
-      'loopback': '^2.22.0',
-      'loopback-component-explorer': '^2.4.0',
+      'loopback': '^2.40.0',
+      'loopback-component-explorer': '^5.4.0',
       'loopback-datasource-juggler': '^2.39.0',
     };
+
     var debug = require('debug')('workspace');
 
     /**
--- a/package.json
+++ b/package.json
@@ -20,10 +20,10 @@
   "dependencies": {
     "async": "^1.5.2",
     "cookie-parser": "^1.3.2",
-    "debug": "^2.2.0",
+    "debug": "^4.0.1",
     "fs-extra": "^0.30.0",
     "glob": "^7.0.0",
-    "lodash": "^4.5.1",
+    "lodash": "^4.17.11",
     "loopback": "^2.0.0",
     "loopback-boot": "^2.6.5",
     "loopback-component-explorer": "^2.1.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "loopback-workspace",
   "version": "3.42.0",
+  "engines": {
+    "node": ">=6 <8"
+  },
   "main": "server/server.js",
   "publishConfig": {
     "export-tests": true,
--- a/templates/projects/empty-server/data.js
+++ b/templates/projects/empty-server/data.js
@@ -23,7 +23,7 @@ template.package = {
   'dependencies': {
     'compression': '^1.0.3',
     'cors': '^2.5.2',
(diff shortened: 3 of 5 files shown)


smartmouse:
we need to pass CI on node 6.x , do not care about node 4 anymore.

common/models/workspace.js
```diff
@@ -36,14 +36,15 @@ module.exports = function(Workspace) {
     var TEMPLATE_DIR = path.join(__dirname, '..', '..', 'templates', 'projects');
     var DEFAULT_TEMPLATE = 'api-server';
     var DEPENDENCIES_3_X = {
-      'loopback': '^3.0.0',
+      'loopback': '^3.22.0',
       'loopback-component-explorer': '^4.0.0',
```
bajtos: Please update explorer to `^6.0.0` while you are making these changes.

common/models/workspace.js
```diff
@@ -36,14 +36,15 @@ module.exports = function(Workspace) {
...
     };
     var DEPENDENCIES_2_X = {
-      'loopback': '^2.22.0',
-      'loopback-component-explorer': '^2.4.0',
+      'loopback': '^2.40.0',
+      'loopback-component-explorer': '^5.4.0',
```
bajtos: DEPENDENCIES_2_X are using versions in Active LTS mode, explorer version 5.x is the right one to use here 👍

bajtos:
> we need to pass CI on node 6.x , do not care about node 4 anymore.

Ideally, we should drop support for Node.js 4.x in `package.json`, but that's out of scope of this pull request.

bajtos:
> > we need to pass CI on node 6.x , do not care about node 4 anymore.

> Ideally, we should drop support for Node.js 4.x in package.json, but that's out of scope of this pull request.

`package.json` is already saying `engines: {node >=6}` in master. I realized just now that this pull request is targeting 3.x branch

bajtos:
@jannyHou `3.x-latest` is a branch created and maintained by `slt-release` tooling. Pull request must target `3.x` branch instead.

bajtos:
I have created the `3.x` branch from the commit 826889da, its the last meaningful change before 4.0 that is clearly not breaking anything. I have also edited the pull request to target `3.x` branch.

I am not sure what GitHub did under the hood, you may need to hard-reset your local git clone instead of doing `git pull`:

```
$ git fetch
$ git checkout update/loopback
$ git reset --hard origin/update/loopback
```

I am not sure what was the major change that caused version bump to 4.0. The only remaining change between 3.x-latest and 4.0 is this explorer-related pull request: https://github.com/strongloop/loopback-workspace/pull/496 You may want to review those changes and decide if you should apply any of them to this pull request, since you are upgrading loopback-explorer too.

I think it would be best to upgrade API Connect to use `loopback-workspace@4`. What are the arguments for staying with `loopback-workspace@3`?

bajtos:
> I am not sure what GitHub did under the hood, you may need to hard-reset your local git clone instead of doing git pull.

PR Linter is saying that PR is out of date and needs to be rebased - maybe you don't need to hard-reset but simply `git rebase 3.x`.

jannyHou:
@bajtos Thank you for the review and created the correct branch!

It is `generator-loopback@2.3.0`(5085 uses it) that uses `loopback-workspace@3.x`.
The minimal version `generator-loopback` using `loopback-workspace@4.x` is 5.x, which is a big jump from 2.3.0. 

I think the version bumped up from 3.x to 4.x to support a flag [`generateOperationScopedModel`](https://github.com/strongloop/loopback-workspace/pull/496/files#diff-18ca367da8eff850776d6d0288bf0e71R24) in explorer, which is related to another flag in `loopback-swagger` that enables `getUpdateOnlyProperties`.

My opinion is if 5080 scaffolds lb3/lb2 app with the old version of explorer, then that flag is not supported anyway, therefore it's fine to not include it in this PR.

bajtos:
> I think the version bumped up from 3.x to 4.x to support a flag generateOperationScopedModel in explorer, which is related to another flag in loopback-swagger that enables getUpdateOnlyProperties.
>
> My opinion is if 5080 scaffolds lb3/lb2 app with the old version of explorer, then that flag is not supported anyway, therefore it's fine to not include it in this PR.

But isn't this pull request changing the project template that is going to be used by 5080 in such way that the new applications will start using the new version of loopback-swagger that includes `getUpdateOnlyProperties` and emit two schemas for each model ("model" and "new model")?

I am not able to fully judge the impact of the changes proposed here, I'll leave it up to you to decide.

Issue #498, opened 2017-10-25
Upgrade workspace from LB2 to LB3


asaf050:
# Description/Steps to reproduce
npm i loopback-workspace

# Expected result
install loopback 3 as a dependency

# Additional information
installing loopback 2 instead

# Acceptance Criteria
- [ ] Follow the [migration guide](https://loopback.io/doc/en/lb3/Migrating-to-3.0.html) to update the LoopBack workspace to use LB3.

ryan-open:
@dhmlau @bajtos Is this intended?

dhmlau:
Thanks for pinging.  No, it's not intended.  With LB will be EOL in April 2019, we should update this. 


Framework | Status | Published | EOL | Runtime | GA | EOL
-- | -- | -- | -- | -- | -- | --
LoopBack 4 | Current | Oct 2018 | Apr 2021(minimum) | Node 10 | Oct 2018 | Apr 2021
Loopback 3 | Active LTS | Dec 2016 | Dec 2019 | Node 8 | Oct 2017 | Dec 2019
Loopback 2 | Maintenance LTS | Jul 2014 | Apr 2019 | Node 6 | Oct 2016 | Apr 2019

I'm reopening this issue.  See if we can make it to Nov (perhaps as stretch goal because currently our Nov sprint is pretty full) or Dec.

dhmlau:
Besides updating to use `loopback@3` by default, I think there are multiple places that should be updated as well.  According to @bajtos, this could be breaking change, so we might as well take this chance to update.

- [ ] update to use `loopback@3` by default
- [ ] Other things to consider
   - [ ] update dependencies of this module, e.g. `strong-globalize` 
   - [ ] update the `engines` in the template to use Node 6+ https://github.com/strongloop/loopback-workspace/blob/master/templates/projects/empty-server/data.js#L17
 

@nabdelgadir, could you please work with @jannyHou to get this task groomed?  Thanks!

PR #539, opened 2018-11-29
Update LB version to LB3

nabdelgadir:
### Description
Update LoopBack version from LoopBack 2 to LoopBack 3.

#### Related issues
- connect to https://github.com/strongloop/loopback-workspace/issues/498

### Checklist

- [ ] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)


.gitignore
```diff
@@ -12,3 +12,5 @@ node_modules/
 /test/sandbox
 /test/.pkgcache/
 dist
+package-lock.json
```
bajtos: We don't use package lock. What to do instead: add `.npmrc` file disabling this feature. You can copy this file e.g. from the `loopback` module, see https://github.com/strongloop/loopback/blob/master/.npmrc

common/models/config-file.json
```diff
@@ -1,5 +1,6 @@
 {
   "validateUpsert": true,
+  "base": "WorkspaceEntity",
```
bajtos: I am not sure whether ConfigFile should be extending WorkspaceEntity. IIIRC, in LoopBack 2.x, models were by default inheriting from a base class that become `PersistedModel` in LB 3.x, and this is the default behavior in 3.x. 

I believe no change should be necessary in this model definition file. Please revert this line.

package.json
```diff
@@ -26,10 +26,10 @@
     "fs-extra": "^0.30.0",
     "glob": "^7.0.0",
     "lodash": "^4.5.1",
-    "loopback": "^2.0.0",
+    "loopback": "^3.23.2",
     "loopback-boot": "^2.6.5",
     "loopback-component-explorer": "^6.0.1",
-    "loopback-datasource-juggler": "^2.27.0",
+    "loopback-connector-mysql": "^5.3.1",
```
bajtos: The workspace API & implementation does not use MySQL database in any means, please remove this dependency.

We do access MySQL database in tests, the MySQL connector is installed by the test suite - no need to do it here.

server/model-config.json
```diff
@@ -11,19 +11,19 @@
   },
   "Definition": {
     "public": false,
-    "dataSource": false
+    "dataSource": "db"
```
bajtos: We are intentionally setting `"dataSource": false`, these models are not persisted by any datasource. Some of them have their own persistence API (e.g. ConfigFile and Workspace), others are "abstract" base classes (models).

Please revert changes in this file too.

server/server.js
```diff
@@ -17,8 +17,7 @@ var errorHandler = require('strong-error-handler');
 
 // required to support base models
 app.dataSource('db', {
-  connector: loopback.Memory,
-  defaultForType: 'db',
+  connector: "memory"
```
bajtos: I think this change should not be necessary, but it should not break anything either, so I'm fine with keeping it :+1:

bajtos:
@nabdelgadir The code base in loopback-workspace is ancient and some files (e.g. `server/server.js`) have not been updated to match the changes in project layout we made since early 1.x days 😞 

Anyhow, here are the lines where to start learning about workspace internals:

https://github.com/strongloop/loopback-workspace/blob/f95230a77758cf33bff7e0f89547f974aac5c2ad/server/server.js#L35-L38

What is happening here: we are modifying `memory` connector in such way that queries are loading data from project files and updates are modifying project files. The implementation details shared by all models are in [server/connector.js](https://github.com/strongloop/loopback-workspace/blob/master/server/connector.js). Then there are entity methods `Entity.allFromCache` &  `Entity.saveToFs` dealing with writing data to the filesystem, and `ConfigFile.findFacetFiles`, `Facet.loadIntoCache` etc. dealing with reading data from the filesystem.

I think the code inside individual model classes (Entity subclasses) should not need any changes to work in LoopBack 3.x. The code inside `server/connector.js` may need small tweaks to make it work with 3.x version of the memory connector, although it's also possible no changes will be needed.

server/server.js
bajtos: Nitpick: please use single quotes (`'memory'`) instead of double quotes.

We use single quotes in JavaScript/TypeScript code. Double quotes are required by JSON standard, thus we use them in JSON files.

bajtos:
ok, the changes made so far looks good to me.

Now we need to find out why the tests are failing.

```
 1) ComponentConfig "before each" hook: givenBasicWorkspace for "should read data from "component-config.json"":
     Error: Cannot call WorkspaceEntity.destroyAll(). The destroyAll method has not been setup. The PersistedModel has not been correctly attached to a DataSource!
```

bajtos:
I see the downstream builds for generator-loopback are still failing. While I think those test failure should not be related to this pull request, we cannot be really sure. 

Let's fix generator-loopback first please. Then trigger a new build and see if all is good.

I opened a pull request to fix generator-loopback tests, see https://github.com/strongloop/generator-loopback/pull/399 Once it's approved & merged, we should back-port it to 5.x branch of generator-loopback too.

cc @jannyHou

bajtos:
@slnode test please

bajtos:
@nabdelgadir all CI checks are green now, yay! 🎉

PR #541, opened 2018-12-10
Update dependencies and node version

nabdelgadir:
Updated dependencies except for the `eslint` and `supertest` dependencies and removed `Gruntfile.js`. 

#### Related issues

- connect to https://github.com/strongloop/loopback-workspace/issues/498

### Checklist

- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)


package.json
```diff
@@ -47,11 +47,10 @@
...
     "eslint-config-loopback": "^2.0.0",
     "grunt": "^1.0.0",
-    "grunt-docular": "~0.1.2",
     "grunt-loopback-sdk-angular": "^1.2.0",
-    "mocha": "^2.4.5",
+    "mocha": "^5.2.0",
     "mysql": "^2.4.2",
     "read": "^1.0.5",
-    "supertest": "^1.2.0"
+    "supertest": "^3.0.0"
```
jannyHou: The error might be caused by the upgrading of supertest.
@nabdelgadir you could try revert the version to `^1.2.0` to narrow down the failing reason. If tests pass, then we can upgrade the api to fit version 3.x

bajtos:
_Cross-posting some information about Grunt from a chat we had elsewhere._

I don't remember why exactly we had grunt & docular in the depencencies. Based on the Gruntfile.js, I think the idea was to generate human-readable documentation for the Angular.js client generated for workspace's REST API. I believe all consumers of those Angular.js clients are no longer actively developed, only in maintenance mode, so it should be ok to remove those grunt-related dependencies.

PR #556, opened 2019-09-30
Remove `npm audit` from scaffolded `posttest` to fix CI builds

bajtos:
New applications are created with a dependency on loopback-component-explorer, which depends on `swagger-ui@2` that's no longer maintained.

This version of swagger-ui has known security vulnerabilities which DO NOT apply to loopback-component-explorer, see the discussion in https://github.com/strongloop/loopback-component-explorer/issues/263.

Unfortunately, there is no way how to tell `npm` to ignore them.

This commit removes `npm audit` from the `posttest` script, to let `npm test` pass and thus fix our failing CI builds.

### Checklist

- [x] New tests added or existing tests modified to cover all changes
- [x] Code conforms with the [style guide](http://loopback.io/doc/en/contrib/style-guide.html)

### DIFF
--- a/templates/projects/empty-server/data.js
+++ b/templates/projects/empty-server/data.js
@@ -26,7 +26,7 @@ template.package = {
   scripts: {
     lint: 'eslint .',
     start: 'node .',
-    posttest: 'npm run lint && npm audit',
+    posttest: 'npm run lint',
   },
   dependencies: {
     compression: '^1.0.3',
--- a/test/end-to-end.js
+++ b/test/end-to-end.js
@@ -1226,7 +1226,7 @@ function localInstall(cwd, cb) {
   const options = {
     cwd: cwd,
   };
-  const script = 'npm install';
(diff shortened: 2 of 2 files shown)


bajtos:
> Are the vulnerabilities only from `swagger-ui@2`? Do we need to detect other vulnerabilities?

Yes, they are from `swagger-ui@2` only. This may change in the future though 🤷‍♂ 

Historically, we were running `nsp check` as part of `posttest` script, the idea was to push LB app developers to deal with security of their dependencies. 

Nowadays, `npm` prints warnings about security vulnerabilities whenever you run `npm install`, I feel that's good enough as far as our project template should be concerned.

bajtos:
> Please squash the commits before merging.

I am intentionally keeping the change split into two commits. If we ever decide to revert this PR and start to run `npm audit` again, we want to revert only the second one, but keep the first.

agnes512:
I see. Thanks for the clarification.

PR #567, opened 2020-09-03
chore: move explorer to devdependency

jannyHou:
Signed-off-by: jannyHou <juehou@ca.ibm.com>

Move explorer from dependency to devDependency to prevent the snyk warning.
See https://github.ibm.com/velox/support/issues/668#issuecomment-23073287

## Checklist

👉 [Read and sign the CLA (Contributor License Agreement)](https://cla.strongloop.com/agreements/strongloop/loopback-workspace) 👈

- [ ] `npm test` passes on your machine
- [ ] Code conforms with the [style guide](https://loopback.io/doc/en/contrib/style-guide-es6.html)
- [ ] Commit messages are following our [guidelines](https://loopback.io/doc/en/contrib/git-commit-messages.html)


server/server.js
```diff
@@ -54,8 +54,19 @@ app.use(methodOverride());
...
-// API explorer
-require('loopback-component-explorer')(app);
+// To prevent the snyk warning about swagger-ui@2.x, the explorer component
+// is moved from dependencies to devDependencies.
+// Here we try to mount the explorer component in case it's not installed.
+
+try {
+  require('loopback-component-explorer')(app);
+} catch (e) {
+  console.log(
```
bajtos, line 64: Please check if the error is caused by the missing explorer package or something else. We have been bitten in the past by code that was hiding legitimate errors and masking them as "module was not found".

You can use the code loading connector modules for inspiration.

https://github.com/strongloop/loopback-datasource-juggler/blob/261dd1c865e8b197882a161b909b90b0e6c3c2a3/lib/datasource.js#L262-L271
```js
      const notFound = e.code === 'MODULE_NOT_FOUND' &&
        e.message && e.message.indexOf(names[m]) > 0;


      if (notFound) {
        debug('Module %s not found, will try another candidate.', names[m]);
        continue;
      }


      debug('Cannot load connector %s: %s', names[m], e.stack || e);
      throw e;
```

bajtos: I would also not use `console.log()`, we don't want to spam users of `lb` and `apic` CLI tools. Can we use `debug()` instead?

bajtos (changes requested):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "loopback-workspace",
-  "version": "5.2.1",
+  "version": "6.0.0",
```
bajtos, line 3: We usually set the version number as part of the release process. Let's use a pre-release version for this PR.

```suggestion
  "version": "6.0.0-dev",
```

bajtos (commented):
Please update also the table with supported versions at the end of the README file, see https://github.com/strongloop/loopback-workspace#module-long-term-support-policy

package.json
jannyHou: @bajtos good catch, I realized apiconnect-cli-loopback depends on 4.x, not 5.x, so we have to patch 4.x instead.

server/server.js
jannyHou: Thanks @bajtos yeah debug is better. And I added the message check for error to destinguish the not found error and others.

jannyHou:
@bajtos Thank you for the detailed suggestions, I also open a PR in https://github.com/strongloop/loopback-workspace/pull/568 to update the 4.x version.

hacksparrow (approved):

server/server.js
```diff
@@ -15,6 +15,7 @@ const methodOverride = require('method-override');
 const app = module.exports = loopback();
 const boot = require('loopback-boot');
 const cookieParser = require('cookie-parser');
+const debug = require('debug')('workspace:server');
```
bajtos, line 18: So far, our convention has been to prefix all loopback-related debug logs with `loopback:` prefix, see e.g. https://github.com/strongloop/loopback/blob/ab2e462ca673cca5deb3c7e3ed2cd3829960d030/lib/model.js#L12

```suggestion
const debug = require('debug')('loopback:workspace:server');
```

server/server.js
```diff
@@ -54,8 +55,24 @@ app.use(methodOverride());
 // LoopBack REST interface
 app.use(app.get('restApiRoot'), loopback.rest());
 
-// API explorer
-require('loopback-component-explorer')(app);
+// To prevent the snyk warning about swagger-ui@2.x, the explorer component
+// is moved from dependencies to devDependencies.
+// Here we try to mount the explorer component in case it's not installed.
```
bajtos, line 60:
```suggestion
// Here we try to mount the explorer component in case it's installed.
```

server/server.js
```diff
@@ -54,8 +55,24 @@ app.use(methodOverride());
...
+
+try {
+  require('loopback-component-explorer')(app);
+} catch (e) {
+  const notFound = e.code === 'MODULE_NOT_FOUND' && e.message;
```
bajtos, line 65: I think you forgot to check that `e.message` is referring to explorer? I think the expression `e.message` is always truthy.

On the second thought, using `require.resolve` may be a simpler option. Quoting from https://nodejs.org/api/modules.html#modules_require_resolve_request_options

> Use the internal require() machinery to look up the location of a module, but rather than loading the module, just return the resolved filename.
> 
> If the module can not be found, a `MODULE_NOT_FOUND` error is thrown.

```js
let hasExplorer;
try {
  require.resolve('loopback-component-explorer');
  hasExplorer = true;
} catch(err) {
  hasExplorer = false;
}

if (hasExplorer) {
  require('loopback-component-explorer')(app);
}
```

server/server.js
```diff
@@ -54,8 +55,24 @@ app.use(methodOverride());
...
+      'loopback-workspace runs without API Explorer enabled.' +
+      'To install it, run `npm i loopback-component-explorer`.',
+    );
+  } else {
+    debug('Cannot load loopback-component-explorer: %s', e.stack || e);
+  }
+}
+
 app.once('started', function(baseUrl) {
   g.log('Browse your REST API at %s%s', baseUrl, '/explorer');
```
bajtos, line 77: This log should not be printed if Explorer is not enabled.

bajtos (changes requested):
The new version is better, I have few more comments to discuss.

server/server.js
jannyHou: @bajtos I think, if I require a certain module and get error "MODULE_NOT_FOUND", then the only possibility is that particular module not found, isn't it?

jannyHou: hmm, true `require.resolve('loopback-component-explorer');` sounds better

server/server.js
jannyHou: good catch.

jannyHou:
@bajtos Appreciate all suggestions!


Start with explorer installed(nothing changed):

```
jannyHous-MBP:loopback-workspace jannyhou$ node .
Browse your REST API at http://0.0.0.0:3000/explorer
LoopBack server listening @ http://0.0.0.0:3000/
```

Uninstall explorer and start with debug:
```
jannyHous-MBP:loopback-workspace jannyhou$ DEBUG=loopback:workspace:server node .
  loopback:workspace:server loopback-workspace runs without API Explorer enabled due to Error: Cannot find module 'loopback-component-explorer'
Require stack:
- /Users/jannyhou/Desktop/August/loopback-workspace/server/server.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:623:15)
    at Function.resolve (internal/modules/cjs/helpers.js:21:19)
    at Object.<anonymous> (/Users/jannyhou/Desktop/August/loopback-workspace/server/server.js:64:11)
    at Module._compile (internal/modules/cjs/loader.js:774:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:785:10)
    at Module.load (internal/modules/cjs/loader.js:641:32)
    at Function.Module._load (internal/modules/cjs/loader.js:556:12)
    at Function.Module.runMain (internal/modules/cjs/loader.js:837:10)
    at internal/main/run_main_module.js:17:11. 
  loopback:workspace:server To install it, run `npm i loopback-component-explorer`. +0ms
LoopBack server listening @ http://0.0.0.0:3000/
```

bajtos (approved):
LGTM 👍🏻

jannyHou:
@slnode test please

